PR #1, opened 2026-04-09
Stamp out exemplar template

mborland:
Generated directly from the bash file in the beman exemplar repo. I left the unlicense, but beman requires either Apache or Boost. They recommend Apache, but Chris and I will likely use this code or derivative code again so we need the BSL.

CC: @ckormanyos


mborland:
I have updated the paper number to reflect P4444. This is good to go from my end.

PR #2, opened 2026-04-09
Fix generation script issues

mborland:


eisenwave (approved):

eisenwave:
We can fix this later, but it should just be `beman::big_int` and `BEMAN_BIG_INT`; `std` shouldn't be in the name of macros or namespaces.

mborland:
The only failure is this in the CML linter:

image

I had to add a `.gersemi` file, otherwise this errored for can't find the beman installer. Hopefully once on main it'll be resolved for the next PR

eisenwave:
Feel free to merge

PR #4, opened 2026-04-09
Add basic construction and ops

mborland:
- Removes the `STD_` and `std_` from namespace and macros like discussed in #2 
- Adds the basic constructors for the `basic_big_int` type which exercise the internal constructors. Tested both at runtime and via `consteval` functions
- Add 3 of the 4 functions from big.int.ops
- Replaces the need to explicitly write out every test in the test CML, now it's just a glob loop

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -5,6 +5,7 @@
 #define BEMAN_BIG_INT_BIG_INT_HPP
 
 #include <algorithm>
+#include <bit>
 #include <climits>
 #include <cstddef>
 #include <cstdint>
@@ -128,6 +129,30 @@ public:
             alloc_traits::deallocate(m_alloc, m_data.ld.data, m_data.ld.capacity);
         }
     }
+
+    // [big.int.ops]
+    [[nodiscard]] constexpr std::size_t width_mag() const noexcept {
+        if (m_limbs == 0) {
+            return 0;
+        }
+
+        const auto top = m_internal ? m_data.la[m_limbs - 1] : m_data.ld.data[m_limbs - 1];
+        return (m_limbs - 1) * bits_per_limb + (bits_per_limb - static_cast<std::size_t>(std::countl_zero(top)) - 1);
+    }
+
+    [[nodiscard]] constexpr std::span<const uint_multiprecision_t> representation() const noexcept {
+        if (m_internal) {
+            return {m_data.la, m_limbs};
+        }
+
+        return {m_data.ld.data, m_limbs};
+    }
+
+    [[nodiscard]] constexpr allocator_type get_allocator() const noexcept {
+        return m_alloc;
+    }
+
+    // constexpr void shrink_to_fit()
 };
 
 template <class Allocator = std::allocator<uint_multiprecision_t>>
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -74,7 +74,9 @@ class basic_big_int {
         limb_type la[inplace_bits];
         limb_data ld;
 
-        constexpr data_type() : la {} {}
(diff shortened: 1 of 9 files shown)


include/beman/big_int/big_int.hpp
```diff
@@ -0,0 +1,161 @@
...
+struct is_basic_big_int<basic_big_int<inplace_bits, Allocator>> : std::true_type {};
+
+template <class T>
+inline constexpr bool is_basic_big_int_v = is_basic_big_int<std::remove_cvref_t<T>>::value;
+
+} // namespace detail
+
+// [big.ing.expos]
+template <class T>
+concept signed_or_unsigned = std::is_signed_v<T> || std::is_unsigned_v<T>;
```
eisenwave, line 45: To be clear, anything in *`kebap-case`* in the paper is exposition-only, so it should go into `detail` or something. I'm not 100% sure what the Beman policy on that is, but I would want those declarations to be separate.

include/beman/big_int/big_int.hpp
```diff
@@ -0,0 +1,161 @@
...
+
+    data_type                            m_data;
+    std::size_t                          m_limbs;    // number of active limbs
+    bool                                 m_sign;     // true = negative (sign + magnitude)
+    bool                                 m_internal; // true = using la[], false = using ld
+    [[no_unique_address]] allocator_type m_alloc;
+
+  public:
+    // [big.int.cons], construct/copy/destroy
+    constexpr basic_big_int() noexcept(noexcept(Allocator())) : m_data{}, m_limbs{}, m_sign{}, m_internal{false} {}
```
eisenwave: We should find a way to start the lifetime of `la` in here. I recall this exact case being rather painful in `std::inplace_vector`, but it should be easy here.

The default constructor should set the integer value to zero, so at least the lowest limb has to be zeroed, and while we're at it, may as well:

```suggestion
    constexpr basic_big_int() noexcept(noexcept(Allocator())) : m_data{.la = {}}, m_limbs{}, m_sign{}, m_internal{true} {}
```

mborland: The default constructor for `m_data` (type `data_type`) is `constexpr data_type() noexcept : la{} {}` So the first comment should already be covered. 

For the change to `m_internal` shouldn't we try to use the static memory for whatever the first operation is rather than defaulting to an allocation?

eisenwave: I thought `m_internal` being `true` means we are using static memory, as documented. That's why I'm surprised you're setting it to `false` (see "false = using ld")

include/beman/big_int/big_int.hpp
```diff
@@ -1,9 +1,161 @@
...
+        }
+
+        const auto top = m_internal ? m_data.la[m_limbs - 1] : m_data.ld.data[m_limbs - 1];
+        return (m_limbs - 1) * bits_per_limb + (bits_per_limb - static_cast<std::size_t>(std::countl_zero(top)) - 1);
+    }
+
+    [[nodiscard]] constexpr std::span<const uint_multiprecision_t> representation() const noexcept {
+        if (m_limbs == 0) {
+            return {};
+        }
```
eisenwave: This case should be impossible because there exists no integer value that would be represented using no limbs. `0` requires one limb to be represented.

Or at least, I'm not sure we ever want to return an empty representation. That seems like a special case for the user to handle.

include/beman/big_int/big_int.hpp
mborland: My mistake, you're correct.

include/beman/big_int/big_int.hpp
mborland: I've updated the default constructor to represent 0 with 1 limb instead of 0 limbs. These checks for no limbs are then removed.

eisenwave (approved):

PR #5, opened 2026-04-09
Rename project files to `big_int` instead of `std-big-int`

eisenwave:


mborland:
Also included in #4. I can rebase on you if you merge first.

PR #6, opened 2026-04-10
Add easy unary operations

eisenwave:

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -147,6 +147,8 @@ class basic_big_int {
 
     [[nodiscard]] constexpr allocator_type get_allocator() const noexcept { return m_alloc; }
 
+    // constexpr void shrink_to_fit()
+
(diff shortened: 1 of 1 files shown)


include/beman/big_int/big_int.hpp
```diff
@@ -148,6 +148,21 @@ class basic_big_int {
...
 
     // constexpr void shrink_to_fit()
+
+    // [big.int.unary]
+    [[nodiscard]] constexpr basic_big_int operator+() const& { return *this; }
+    [[nodiscard]] constexpr basic_big_int operator+() && noexcept { return std::move(*this); }
+
+    [[nodiscard]] constexpr basic_big_int operator-() const& {
+        auto copy = *this;
+        copy.m_sign ^= 1;
```
mborland, line 158: GCC and Clang will both emit warning `-Wbool-operation` here. Could be changed safely to `copy.m_sign = !copy.m_sign`

PR #7, opened 2026-04-10
Add various utilities and wide operations

eisenwave:

### DIFF
--- a/include/beman/big_int/wide_ops.hpp
+++ b/include/beman/big_int/wide_ops.hpp
@@ -1,0 +1,134 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-License-Identifier: BSL-1.0
+
+#ifndef BEMAN_BIG_INT_WIDE_OPS_HPP
+#define BEMAN_BIG_INT_WIDE_OPS_HPP
+
+#include <limits>
+#include <bit>
+
+#include <beman/big_int/config.hpp>
+
+namespace beman::big_int::detail {
+
+template <std::integral T>
+inline constexpr std::size_t width_v = std::numeric_limits<std::make_unsigned_t<T>>::digits;
+
+struct wide {
+    uint_multiprecision_t low_bits;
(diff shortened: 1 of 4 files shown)


include/beman/big_int/wide_ops.hpp
```diff
@@ -0,0 +1,210 @@
...
+template <signed_or_unsigned T>
+[[nodiscard]] constexpr overflow_result<T> overflowing_mul(T x, T y) noexcept {
+#ifdef BEMAN_BIG_INT_GNUC
+    T    value;
+    bool overflow = __builtin_mul_overflow(x, y, &value);
+    return {.value = value, .overflow = overflow};
+#else
+    static_assert(width_v<T> <= 64, "Don't need more than 64-bit checked operations for now.");
+    using Wide = std::conditional_t<std::is_signed_v<T>, int128_t, uint128_t>;
+    Wide wide  = static_cast<Wide>(x) * static_cast<Wide>(y);
```
mborland, line 128: Small nit pick on these ops. 128 * 64 is usually an optimized path vs 128 * 128. Same for the others.

eisenwave: I think it's fine to crutch on modern compilers. GCC and Clang emit a single `mul` instruction here: https://godbolt.org/z/fbczMcGr7

The compiler can recognize that the 128-bit multiplication has only the lower 64 bits set in each operand and can figure out how to best lower that.

mborland: I was thinking about the MSVC case since they use Knuth for multiplication in some cases as they lack the `__builtin_mul_overflow` intrinsic. On 64-Bit this should just become `_umul128` though.

eisenwave: You're right, the codegen doesn't look great for MSVC, but it doesn't look great even with `_umul128`: https://godbolt.org/z/oGbssW7b5

That's more of an issue with calling conventions though; the returned `struct` is not passed by register pair.

PR #8, opened 2026-04-10
Add integral and ranges constructors

ghost:


include/beman/big_int/big_int.hpp
```diff
@@ -47,6 +48,13 @@ concept signed_or_unsigned = std::is_signed_v<T> || std::is_unsigned_v<T>;
 template <class T>
 concept arbitrary_integer = signed_or_unsigned<std::remove_cvref_t<T>> || detail::is_basic_big_int_v<T>;
 
+template <std::size_t inplace_bits, class T>
+inline constexpr bool no_alloc_constructible_from = sizeof(std::remove_cvref_t<T>) * CHAR_BIT <= inplace_bits;
```
eisenwave: This should be rebased on #7. I'll merge that one in a sec. After doing so, this should be
```suggestion
inline constexpr bool no_alloc_constructible_from = width_v<std::remove_cvref_t<T>> <= inplace_bits;
```

include/beman/big_int/big_int.hpp
```diff
@@ -116,6 +124,43 @@ class basic_big_int {
         : m_data{}, m_limbs{1}, m_sign{}, m_internal{true}, m_alloc{a} {}
     constexpr basic_big_int(const basic_big_int& x)     = default;
     constexpr basic_big_int(basic_big_int&& x) noexcept = default;
+    template <detail::signed_or_unsigned T>
+    constexpr explicit(!detail::is_implicit_constructible_from<inplace_bits, Allocator, T>)
+        basic_big_int(T value) noexcept(detail::no_alloc_constructible_from<inplace_bits, T>)
```
eisenwave: In the paper, this is constrained on *`arbitrary-arithmetic-type`* so this would also include construction from floating-point types. There should be a TODO somewhere here saying that the constructor is incomplete. Also, it shouldn't be taking by value.

mborland:
@alcxpr I rebased you and addressed Jan's comments. I added in my initial thought as a comment on how to implement the floating point constructor.

eisenwave (approved):

alcxpr:
Thank you very much for the fix. Will be doing what you suggest regarding the Ryu FP construction.

mborland:
There's a complete Ryu in boost charconv that can be lifted out: https://github.com/boostorg/charconv/tree/develop/include/boost/charconv/detail/ryu

PR #18, opened 2026-04-10
Add floating-point support for constructors

ghost:
Do note that I didn't write tests for FP types in `<stdfloat>`  because there's no compiler support for it.


include/beman/big_int/big_int.hpp
```diff
@@ -479,6 +578,85 @@ constexpr void basic_big_int<inplace_bits, Allocator>::assign_magnitude(T value)
...
+    constexpr int mb   = traits::mantissa_bits;
+    constexpr int bias = traits::bias;
+
+    bits_t        bits;
+    std::uint32_t ieee_exp;
+    std::uint64_t ieee_mantissa;
+
+#if __LDBL_MANT_DIG__ == 64 && __LDBL_MAX_EXP__ == 16384
+    if constexpr (std::is_same_v<bits_t, detail::long_double_bits>) {
+        __builtin_memcpy(&bits, &value, sizeof(bits));
```
eisenwave, line 512: I don't think this works during constant evaluation, does it? `std::bit_cast` should work.

include/beman/big_int/big_int.hpp
```diff
@@ -46,12 +49,103 @@ template <class T>
...
+        return false;
+    }
+}();
 
 template <std::size_t inplace_bits, class Allocator, class T>
 inline constexpr bool is_implicit_constructible_from =
     detail::signed_or_unsigned<std::remove_cvref_t<T>> ||
     std::is_same_v<std::remove_cvref_t<T>, basic_big_int<inplace_bits, Allocator>>;
+
+template <class F>
```
eisenwave: These traits should really go in a separate header.

tests/beman/big_int/float_construction.test.cpp
```diff
@@ -0,0 +1,61 @@
...
+    GTEST_SKIP() << "long double not supported on this platform";
+#else
+    beman::big_int::big_int x(42.9L);
+    EXPECT_EQ(x.representation()[0], 42U);
+#endif
+}
+
+TEST(FloatConstruction, DoubleLargeMultiLimb) {
+    beman::big_int::big_int x(static_cast<double>(1ULL << 63) * 4.0);
+    EXPECT_EQ(x.representation().size(), 2U);
```
eisenwave: This test could be more detailed.

We can hold off on that until we have `operator==` though.

eisenwave (commented):
If it works, it's good to merge for now, but there's some follow-up work to be done.

include/beman/big_int/big_int.hpp
alcxpr: Wouldn't using `std::bit_cast` here be UB on x86 due to the 6 padding bytes? I think [P3969](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2026/p3969r0.html) said something about it.

We probably can leave constant evaluation case as

```cpp
if consteval {
    return;
}
```
inside the constexpr if for now.

PR #21, opened 2026-04-11
Fix various allocation-related problems

eisenwave:
This fixes a bunch of problems:
- Various places use raw pointers instead of the allocator's fancy `pointer`.
- `big_int` does rebinding, but allocator-aware containers like `vector` typically don't do that.
- `allocate_at_least` isn't used, which is wasteful.
- Unnecessary allocation in copy constructor.
- Unnecessary and weird special casing in `constexpr` which bypasses `allocate` via `new` expression.
- Undefined behavior due to `std::copy_n` to uninitialized storage (must be `uninitialized_copy_n`)

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -58,6 +58,17 @@ template <std::size_t inplace_bits, class Allocator, class T>
 inline constexpr bool is_implicit_constructible_from =
     detail::signed_or_unsigned<std::remove_cvref_t<T>> ||
     std::is_same_v<std::remove_cvref_t<T>, basic_big_int<inplace_bits, Allocator>>;
+
+#if __cpp_lib_allocate_at_least >= 202302L
+using std::allocation_result;
+#else
+template <class Pointer, class SizeType = std::size_t>
+struct allocation_result {
+    Pointer  ptr;
+    SizeType count;
+};
+#endif // __cpp_lib_allocate_at_least
+
 } // namespace detail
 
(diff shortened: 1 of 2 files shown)


include/beman/big_int/big_int.hpp
```diff
@@ -58,6 +58,17 @@ template <std::size_t inplace_bits, class Allocator, class T>
 inline constexpr bool is_implicit_constructible_from =
     detail::signed_or_unsigned<std::remove_cvref_t<T>> ||
     std::is_same_v<std::remove_cvref_t<T>, basic_big_int<inplace_bits, Allocator>>;
+
+#if __cpp_lib_allocate_at_least >= 202302L
```
mborland, line 62:
```suggestion
#if defined(__cpp_lib_allocate_at_least) && __cpp_lib_allocate_at_least >= 202302L
```

I recommend this form for feature test macros to avoid `-Wundef` warnings.

mborland:
Thanks for the fixes, they all look good. I didn't know about `allocate_at_least` so that's good to know now.

include/beman/big_int/big_int.hpp
eisenwave: Can you just pepper the presets with strict warnings so that things won't build in the first place? `-Wall -Wextra -Wpedantic -Wconversion -Wsign-conversion -Wundef ...`

mborland: See: https://github.com/eisenwave/std-big-int/pull/24. Includes all the flags I normally use.

PR #22, opened 2026-04-11
Clean up inplace_bits and signatures

eisenwave:
- Differentiate between `min_inplace_bits` (provided as a template argument, and rounded up) and `inplace_bits` (computed in the instantiation and accessible from the outside).
- Replace `inplace_bits` and `Allocator` in function signatures with `b` and `A`. The `allocator_type` and `inplace_bits` members should be used instead.
- Reduce the default `min_inplace_bits` from `128` to `64`. Following #13, 128 bits is no longer necessary.
- Add various fixes to compile with GCC 13 (feature testing).
- Fix `[[no_unique_address]]` being ignored by MSVC (use `[[msvc::no_unique_address]]`).

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -24,7 +24,7 @@ namespace beman::big_int {
 using beman::big_int::uint_multiprecision_t;
 
 // Forward decl so that we can define our concepts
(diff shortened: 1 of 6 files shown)


mborland:
This also looks good to me

PR #23, opened 2026-04-11
Comparison operators

eisenwave:


eisenwave:
Just needs testing, but should otherwise be good to go now.

PR #38, opened 2026-04-12
Trivial syntax arm-none-eabi warn zero nullptr

ckormanyos:
Hi @eisenwave thx for working so hard on the 32/64 limb topic. I took a run at it on `arm-none-eabi-gcc` with many warnings activated.

This simplistic PR is intended to handle a final syntax warning (warn zero as nullptr) on `arm-none-eabi-gcc`

Cc: @eisenwave

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -892,16 +892,16 @@ constexpr std::strong_ordering operator<=>(const L& lhs, const R& rhs) noexcept 
         }
     } else {
         static_assert(detail::is_basic_big_int_v<R>);
-        #if (defined(BEMAN_BIG_INT_GCC) || defined(BEMAN_BIG_INT_CLANG))
+#if (defined(BEMAN_BIG_INT_GCC) || defined(BEMAN_BIG_INT_CLANG))
         BEMAN_BIG_INT_DIAGNOSTIC_PUSH()
         BEMAN_BIG_INT_DIAGNOSTIC_IGNORED("-Wzero-as-null-pointer-constant")
-        #endif
+#endif
         static_assert((0 <=> std::strong_ordering::less) == std::strong_ordering::greater,
                       "This trick to flip the ordering should work.");
         return 0 <=> rhs.compare_integer(lhs);
-        #if (defined(BEMAN_BIG_INT_GCC) || defined(BEMAN_BIG_INT_CLANG))
+#if (defined(BEMAN_BIG_INT_GCC) || defined(BEMAN_BIG_INT_CLANG))
         BEMAN_BIG_INT_DIAGNOSTIC_POP()
-        #endif
+#endif
     }
 }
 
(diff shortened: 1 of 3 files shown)


ckormanyos:
Well CI was a disaster. I can repair it locally. I am using a specialized allocator, so maybe the error is on my side?

ckormanyos:
>  I am using a specialized allocator, so maybe the error is on my side?

I guess, no matter what allocator I use, I get (on `gcc-arm-none-eabi` 15) a warning as shown below.

```
1>In file included from src/app/benchmark/app_benchmark_std_big_int.cpp:14:
1>C:/ChrisGitRepos/std-big-int/include/beman/big_int/big_int.hpp: In function 'constexpr std::strong_ordering beman::big_int::operator<=>(const L&, const R&)':
1>C:/ChrisGitRepos/std-big-int/include/beman/big_int/big_int.hpp(665,24): warning G1F1615EE: zero as null pointer constant [-Wzero-as-null-pointer-constant]
1>  665 |         static_assert((0 <=> std::strong_ordering::less) == std::strong_ordering::greater,
1>      |    
```

eisenwave:
I think the reason why this happens is that `std::strong_ordering` is specified to be comparable with the integer literal 0, and I guess compilers implement this by providing a comparison operator that takes `const void*` (where `0` can be used as a null pointer constant). However, that technique isn't used everywhere.

I think you should suppress the warning or write out the `0 <=> std::strong_ordering::less` trick using some if statements or conditional operator.

eisenwave (approved):

include/beman/big_int/big_int.hpp
```diff
@@ -892,9 +892,16 @@ constexpr std::strong_ordering operator<=>(const L& lhs, const R& rhs) noexcept
         }
     } else {
         static_assert(detail::is_basic_big_int_v<R>);
+#if (defined(BEMAN_BIG_INT_GCC) || defined(BEMAN_BIG_INT_CLANG))
+        BEMAN_BIG_INT_DIAGNOSTIC_PUSH()
+        BEMAN_BIG_INT_DIAGNOSTIC_IGNORED("-Wzero-as-null-pointer-constant")
+#endif
```
eisenwave: Please don't put this diagnostic macros into `#if`; that only makes the code more cluttered. 
```suggestion
    BEMAN_BIG_INT_DIAGNOSTIC_PUSH()
    BEMAN_BIG_INT_DIAGNOSTIC_IGNORED_GCC("-Wzero-as-null-pointer-constant")
    BEMAN_BIG_INT_DIAGNOSTIC_IGNORED_CLANG("-Wzero-as-null-pointer-constant")
```

include/beman/big_int/big_int.hpp
```diff
@@ -1018,9 +1025,16 @@ constexpr std::strong_ordering basic_big_int<b, A>::compare_integer(const Intege
         if constexpr (has_inplace_to_bit_uint) {
             if (is_storage_static()) {
                 const auto sign_compare = (x < 0) <=> is_negative();
+#if (defined(BEMAN_BIG_INT_GCC) || defined(BEMAN_BIG_INT_CLANG))
+                BEMAN_BIG_INT_DIAGNOSTIC_PUSH()
+                BEMAN_BIG_INT_DIAGNOSTIC_IGNORED("-Wzero-as-null-pointer-constant")
+#endif
                 if (sign_compare != 0) {
```
eisenwave: None of this needs suppression if you write it like:
```suggestion
                if (std::is_neq(sign_compare)) {
```
I would prefer that rather than adding many lines of diagnostic management.

eisenwave (changes requested):

include/beman/big_int/big_int.hpp
ckormanyos: Agreed. I will give it a try. Thank you

include/beman/big_int/big_int.hpp
ckormanyos: Oh that's cool. Thanks

ckormanyos:
Hi Jan (@eisenwave), this one is getting close. I still need to handle your review comments fully.

But there seem to be some unrelated MSVC test errors that are not present in the main branch. I don't see how the changes I made could be causing them.

Also I found something on my 32-bit bare-metal embedded target. My compiler errors out on lines like [lines 97-100 here](https://github.com/eisenwave/std-big-int/blob/4e19c65141e7fa87c12b8c8b5c048929bb456585/include/beman/big_int/wide_ops.hpp#L97). My compiler claims that the variables `a` and `b` cant't be found. Are they supposed to be `x` and `y`?

eisenwave (approved):
Not sure what you mean by the MSVC errors; the build is green now. If you mean the null pointer stuff, I guess we just didn't catch that with warnings in our CI. Not sure why there would be a difference between our CI MSVC though.

`a` and `b` were supposed to be `x` and `y`, yeah.

ckormanyos:
Hi Jan (@eisenwave) do you prefer-squach-and-merge? Or do you want the full details of the long patwhay to each PR?

eisenwave:
I've disabled all methods other than squash for merging a few days ago anyway.

PR #41, opened 2026-04-13
Fix buffer overflow

### LINKED ISSUE #27: Buffer overflow for unsized ranges
https://github.com/eisenwave/std-big-int/blob/5afe0f1d99aa02d3f2ed53fe1e01534f410ffaf1/include/beman/big_int/big_int.hpp#L415-L440

In this part, if we didn't reserve (and we only do so for sized ranges), then we simply write past the end of our storage.

This part is surprisingly difficult to do properly. What needs to happen in the unsized case is:
- Try to write as many limbs to the storage that's already available. If the end of the range is hit before we run out, all is good.
- If we hit the end of our storage (be that inplace or dynamic), we grow, and do so exponentially many times as needed until we can fit all limbs from the range.

ghost:
Fix #27.

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -510,8 +510,8 @@ constexpr basic_big_int<b, A>::basic_big_int(std::from_range_t, R&& r, const all
 
     if constexpr (std::ranges::sized_range<R>) {
         reserve(std::ranges::size(r));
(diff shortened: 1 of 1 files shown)


alcxpr:
Seems to be working now I think, since the sanitizers didn't scream during the tests in CI.

PR #43, opened 2026-04-13
Add f32 and f64 to u128 conversion fast path

### LINKED ISSUE #29: Use `float32_t` -> `uint128_t` conversions where possible
It just so happens that the greatest finite value of a float fits into an unsigned 128-bit integer. That conversions corresponds to the `__fixunssfti` runtime library function or to `fptoui float %0 to i128` in LLVM. This intrinsic functionality can be utilized in assignment from `float` or in the floating-point constructor.

For `float64_t`, the conversion requires `unsigned _BitInt(1024)` and I'm not sure if it's really worth it, but for `float`, this should be the best possible conversion (at least for code size).

ghost:
Fix #29.

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -1168,6 +1168,30 @@ constexpr void basic_big_int<b, A>::assign_from_float(const F value) noexcept {
         return;
     }
 
+#if defined(BEMAN_BIG_INT_HAS_BITINT) && BEMAN_BIG_INT_BITINT_MAXWIDTH >= 128
+    if constexpr (std::is_same_v<F, float>) {
+        // note: not using `assign_magnitude` here since the concept doesn't accept `_BitInt` types.
+        // also `FLT_MAX` needs 2 limbs on a 64-bit system, so we do need `grow(2)` first.
+        grow(2);
+        const auto mag   = static_cast<detail::uint128_t>(value < 0.0f ? -value : value);
+        limb_ptr()[0]    = static_cast<limb_type>(mag);
+        limb_ptr()[1]    = static_cast<limb_type>(mag >> bits_per_limb);
+        const auto count = limb_ptr()[1] != 0 ? 2u : 1u;
+        set_limb_count(static_cast<std::uint32_t>(count));
+        return;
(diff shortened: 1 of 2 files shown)


include/beman/big_int/big_int.hpp
```diff
@@ -1168,6 +1168,30 @@ constexpr void basic_big_int<b, A>::assign_from_float(const F value) noexcept {
         return;
     }
 
+#if defined(BEMAN_BIG_INT_HAS_BITINT) && BEMAN_BIG_INT_BITINT_MAXWIDTH >= 128
```
eisenwave, line 1171: `_BitInt` shouldn't be a prerequisite here. This should work with any 128-bit integer, so `BEMAN_BIG_INT_HAS_INT128`.

include/beman/big_int/big_int.hpp
```diff
@@ -1168,6 +1168,30 @@ constexpr void basic_big_int<b, A>::assign_from_float(const F value) noexcept {
...
+    if constexpr (std::is_same_v<F, float>) {
```
eisenwave, line 1172: This should also work with `std::float32_t` if available.

include/beman/big_int/big_int.hpp
```diff
@@ -1168,6 +1168,30 @@ constexpr void basic_big_int<b, A>::assign_from_float(const F value) noexcept {
...
+        // note: not using `assign_magnitude` here since the concept doesn't accept `_BitInt` types.
+        // also `FLT_MAX` needs 2 limbs on a 64-bit system, so we do need `grow(2)` first.
```
eisenwave, line 1174: `detail::signed_or_unsigned` should be changed to accept bit-precise integer types as well. In libc++, I think that works out of the box, but in libstdc++ and MSVC STL, we'd need to not rely on `std::is_integral_v` alone.

include/beman/big_int/big_int.hpp
alcxpr: I'll make a quick fix on these as a separate PR ASAP.

include/beman/big_int/big_int.hpp
alcxpr: I think I'll make a separate issue for this.

PR #45, opened 2026-04-13
Add std::float32 and std::float64 support and better conditional comp…

ghost:
This partially addresses the code review in #43.

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -1227,12 +1227,24 @@ constexpr void basic_big_int<b, A>::assign_from_float(const F value) noexcept {
     #endif
         ;
 
-    const auto assign_via_uint128 = [&](auto v) {
+const auto assign_via_uint128 = [&](auto v) {
         grow(2);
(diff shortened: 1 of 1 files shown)


include/beman/big_int/big_int.hpp
```diff
@@ -1252,33 +1272,15 @@ constexpr void basic_big_int<b, A>::assign_from_float(const F value) noexcept {
...
             assign_via_uint128(value);
-            return;
+        } else {
+            assign_general();
         }
+    } else {
+        assign_general();
     }
+#else
+    assign_general();
```
eisenwave, line 1282: Do you really need three `else` cases with `assign_general`? I think you could avoid the lambda and just let control flow into it rather than putting it in an `else` statement.

alcxpr: MSVC outputs unreachable code error when the code is what you described, like in my previous commits.

I am not exactly sure how to avoid that but this seems to work.

eisenwave (approved):

PR #46, opened 2026-04-13
Implement operators add and sub

mborland:
- Implements operators add and sub through unified generating function
- Add NOOP `uabs` function
- Add `integer_signbit` helpers
- Pull out magnitude logic from `operator<=>` for use in add and sub
- Add MSVC path for `carrying_add` and `borrowing_sub`
- Fix MSVC detection macro for `widening_mul`

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -972,6 +972,46 @@ template <class T, std::size_t N>
     return std::span<const T, N>(arr);
 }
 
+// Three-way compares the magnitudes represented by two limb spans (little-endian, zero-padded).
+// Treats both as non-negative; callers layer sign handling on top.
+// Split on which side is longer so each branch carries only one high-tail scan.
+// When an extent is not `dynamic_extent`, `.size()` is a compile-time constant,
+// so the loops are easier to unroll.
+template <std::size_t extent_a, std::size_t extent_b>
+[[nodiscard]] constexpr std::strong_ordering
+compare_limb_magnitudes(const std::span<const uint_multiprecision_t, extent_a> a,
+                        const std::span<const uint_multiprecision_t, extent_b> b) noexcept {
+    if (a.size() > b.size()) {
+        // If there are more significant nonzero digits in `a`, it is greater.
+        // Decimal example: 123 > 23
+        for (std::size_t i = a.size(); i-- > b.size();) {
+            if (a[i] != 0) {
+                return std::strong_ordering::greater;
+            }
+        }
+        // Compare the common digits from most to least significant.
+        for (std::size_t i = b.size(); i-- > 0;) {
+            const auto result = a[i] <=> b[i];
+            if (std::is_neq(result)) {
+                return result;
+            }
+        }
+    } else {
+        for (std::size_t i = b.size(); i-- > a.size();) {
+            if (b[i] != 0) {
+                return std::strong_ordering::less;
+            }
+        }
+        for (std::size_t i = a.size(); i-- > 0;) {
+            const auto result = a[i] <=> b[i];
+            if (std::is_neq(result)) {
+                return result;
+            }
+        }
+    }
+    return std::strong_ordering::equal;
+}
+
 } // namespace detail
 
 template <std::size_t b, class A>
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -1277,10 +1277,10 @@ basic_big_int<b, A>::assign_sum_of_limbs(const std::span<const uint_multiprecisi
     // When `lhs >= rhs` (in magnitude) compute `lhs - rhs` and take sign `lhs_neg`; otherwise
     // compute `rhs - lhs` with sign `rhs_neg`. Equal magnitudes fall into the first branch
     // and produce a normalized `+0`.
-    const std::span<const uint_multiprecision_t> larger  = std::is_gteq(magnitude_order) //
+    const std::span<const uint_multiprecision_t> larger  = std::is_gteq(magnitude_order)
                                                               ? std::span<const uint_multiprecision_t>(lhs)
                                                               : std::span<const uint_multiprecision_t>(rhs);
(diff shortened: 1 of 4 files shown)


include/beman/big_int/big_int.hpp
```diff
@@ -288,6 +306,12 @@ class BEMAN_BIG_INT_TRIVIAL_ABI basic_big_int {
     template <class L, detail::common_big_int_type_with<L> R>
     friend constexpr std::strong_ordering operator<=>(const L& lhs, const R& rhs) noexcept;
 
+    // [big.int.binary]
+    template <class L, detail::common_big_int_type_with<L> R>
+    friend constexpr detail::common_big_int_type<L, R> operator+(L&& x, R&& y);
+    template <class L, detail::common_big_int_type_with<L> R>
+    friend constexpr detail::common_big_int_type<L, R> operator-(L&& x, R&& y);
```
eisenwave: I don't think the type-constraint on the template parameters is required. SFINAE kicks in because `common_big_int_type` appears in the return type anyway.

include/beman/big_int/wide_ops.hpp
```diff
@@ -271,6 +271,42 @@ template <unsigned_integer T>
...
     unsigned long long value = __builtin_addcll(x, y, carry, &carry_out);
     return {.value = value, .carry = carry_out != 0};
+#elif defined(BEMAN_BIG_INT_MSVC) && (defined(_M_AMD64) || defined(_M_IX86))
+    // Theoretically we could use the ADX intrinsic, but then the user always has to compile for it
+    // This is more portable.
+    // Each branch returns directly, so we don't share uninitialized state
+    // across the `if constexpr` chain.
+    // We also have to duplicate the fall through case because if not consteval is broken on msvc
+
+    if consteval {
```
eisenwave:
```suggestion
    if BEMAN_BIG_INT_IS_CONSTEVAL {
```

include/beman/big_int/wide_ops.hpp
```diff
@@ -296,6 +332,40 @@ template <unsigned_integer T>
     unsigned long long borrow_out;
     unsigned long long value = __builtin_subcll(x, y, borrow, &borrow_out);
     return {.value = value, .borrow = borrow_out != 0};
+#elif defined(BEMAN_BIG_INT_MSVC) && (defined(_M_AMD64) || defined(_M_IX86))
+    // Mirror the `carrying_add` MSVC path using the matching `_subborrow_*` intrinsics.
+    // Each branch returns directly, so we don't share uninitialized state across the
+    // `if constexpr` chain.
+
+    if consteval {
```
eisenwave:
```suggestion
    if BEMAN_BIG_INT_IS_CONSTEVAL {
```

include/beman/big_int/big_int.hpp
```diff
@@ -331,6 +355,15 @@ class BEMAN_BIG_INT_TRIVIAL_ABI basic_big_int {
...
                                                                bool limbs_negative) const noexcept;
 
+    // Overwrites `*this` with the sum `(lhs, lhs_neg) + (rhs, rhs_neg)`,
+    // Assumes `*this` is in the default-constructed state (inline storage, magnitude zero, sign positive).
+    // Only allocates heap storage if the final result is certain to exceed the inline limb capacity
+    template <std::size_t extent_a, std::size_t extent_b>
+    constexpr void assign_sum_of_limbs(std::span<const uint_multiprecision_t, extent_a> lhs,
+                                       bool                                             lhs_neg,
+                                       std::span<const uint_multiprecision_t, extent_b> rhs,
+                                       bool                                             rhs_neg);
```
eisenwave: This design seems very clunky. If the function only works if you start out with a default constructed object, why don't you just make it `static` and initialize a fresh `basic_big_int` inside?

include/beman/big_int/big_int.hpp
```diff
@@ -954,8 +987,106 @@ template <class T, std::size_t N>
...
+template <class L, detail::common_big_int_type_with<L> R>
+constexpr detail::common_big_int_type<L, R> operator+(L&& x, R&& y) {
+    using Result = detail::common_big_int_type<L, R>;
+    using LT     = std::remove_cvref_t<L>;
+    using RT     = std::remove_cvref_t<R>;
+
+    Result result;
+
+    if constexpr (detail::is_basic_big_int_v<LT> && detail::is_basic_big_int_v<RT>) {
+        result.assign_sum_of_limbs(x.representation(), x.is_negative(), y.representation(), y.is_negative());
```
eisenwave: With the approach in the comment above, this would be
```cpp
return make_sum_of_limbs(x.representation(), x.is_negative(), y.representation(), y.is_negative());
```

include/beman/big_int/big_int.hpp
```diff
@@ -954,8 +987,106 @@ template <class T, std::size_t N>
...
+        }
+    }
+    return std::strong_ordering::equal;
+}
+
 } // namespace detail
 
+// [big.int.binary]
+template <class L, detail::common_big_int_type_with<L> R>
+constexpr detail::common_big_int_type<L, R> operator+(L&& x, R&& y) {
```
eisenwave, line 1035: We are missing optimization opportunities here, but those could be added later.

Namely, if `L` or `R` are non-`const` rvalues, this could be turned into `x += y` or `y += x` and perform the operation in-place using the already existing allocation.

That can be a follow-up improvement though.

eisenwave (changes requested):

include/beman/big_int/wide_ops.hpp
mborland: I see you added `BOOST_BIG_INT_IS_NOT_CONSTEVAL` so I'll use that and remove the duplication that was the workaround.

include/beman/big_int/big_int.hpp
mborland: Seems to work fine without

include/beman/big_int/big_int.hpp
mborland: Tracked as #48

eisenwave (approved):

PR #47, opened 2026-04-13
Switch vars ab to xy in M64 arch path

ckormanyos:
I believe the same switch-up of `ab` to ` xy` is needed in this architecture branch. I hope I got that right. I wonder why CI is not hitting it? Hi @mborland can we devise a test to at least compile this architecture branch?


mborland:
It's not being hit because the macro is wrong. It should be _M_AMD64. I have that fix in the add branch if you want to cherry pick it

mborland:
Cherry picked for you

eisenwave (approved):

ckormanyos:
Thanks Jan. Thanks Matt

PR #49, opened 2026-04-13
Non-const-ify a changed func param

ckormanyos:
The function parameter `const T value` was initially supplied as const, but then shifted in the subroutine.

The test passed, just the `const`-ness of the parameter needs to be removed

Found on bare-metal target.

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -1133,7 +1133,7 @@ basic_big_int<b, A>::compare_limbs(const std::span<const uint_multiprecision_t, 
 
 template <std::size_t b, class A>
 template <detail::unsigned_integer T>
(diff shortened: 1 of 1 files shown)


ckormanyos:
For completeness, here is the pseudo-code of the test used to find the param.

```
  my_big_int_type my_big_int_3 { (std::numeric_limits<local_limb_type>::max)() };

  ++my_big_int_3;

  const bool
    result_ovf_is_ok
    {
         (my_big_int_3.representation()[std::size_t { 0U }] == local_limb_type { 0U })
      && (my_big_int_3.representation()[std::size_t { 1U }] == local_limb_type { 1U })
    };

  result_is_ok = (result_ovf_is_ok && result_is_ok );
```

PR #54, opened 2026-04-14
Reduce unnecessary allocations

### LINKED ISSUE #26: `operator=` is wasteful with allocation
https://github.com/eisenwave/std-big-int/blob/5afe0f1d99aa02d3f2ed53fe1e01534f410ffaf1/include/beman/big_int/big_int.hpp#L335-L358

The problematic part is the unconditional `free_storage()` call at the start. It's theoretically possible that we already have an allocation large enough to hold the value of `x`. However, this code throws away that allocation and makes a new one for no good reason in that case.

### LINKED ISSUE #48: Add and Sub optimizations
From @eisenwave in the original PR:

If L or R are non-const rvalues, this could be turned into x += y or y += x and perform the operation in-place using the already existing allocation.

I'm sure there are other optimization possibilities

mborland:
This PR addresses the allocation issues in add, sub, and `operator=`. We now have a combined `copy_value` function that attempts to reuse storage if it's big enough, or allocates the correct amount as well as any amount of extra limbs. In the add case we request an extra limb each time, except for when we are already using both, or the maximum amount in the general case, of limbs of our in-place storage. In that case we defer that decision to the end of ripple carry at which point we either fit or we have no choice and have to allocate and move ourselves onto the heap.

Closes: #26 
Closes: #48

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -1040,13 +1040,11 @@ compare_limb_magnitudes(const std::span<const uint_multiprecision_t, extent_a> a
 // constructing from a primitive), then fold the other operand in via `add_in_place`.
 // `add_in_place` handles carry, borrow, trim, and sign normalization uniformly.
 //
-// Dispatch priority for the destination is:
-//   1. lhs is an rvalue `basic_big_int`   -> move-from lhs
-//   2. rhs is an rvalue `basic_big_int`   -> move-from rhs (addition is commutative;
-//                                             for subtraction we negate first)
-//   3. lhs is an lvalue `basic_big_int`   -> copy lhs
-//   4. otherwise lhs is a primitive and rhs must be an lvalue `basic_big_int`
-//                                          -> copy rhs (with negate for subtraction)
+// Destination priority:
+//   * both `basic_big_int` rvalues  -> move the one with more limbs (no subsequent grow)
+//   * one  `basic_big_int` rvalue   -> move it
+//   * both `basic_big_int` lvalues  -> copy the one with more limbs
+//   * one  `basic_big_int` lvalue   -> copy it (the other side is a primitive)
 //
 // `common_big_int_type` only yields a type when both `basic_big_int` operands are the
 // exact same `basic_big_int<b, A>` instantiation, so the operand type already matches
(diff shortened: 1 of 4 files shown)


include/beman/big_int/big_int.hpp
```diff
@@ -1035,65 +1099,169 @@ compare_limb_magnitudes(const std::span<const uint_multiprecision_t, extent_a> a
...
+        return r;
+    } else if constexpr (detail::is_basic_big_int_v<LT> && detail::is_basic_big_int_v<RT>) {
+        // 4) both lvalue `basic_big_int`s: `copy_value` folds the copy of the
+        // larger operand and the carry-headroom reservation into one allocation
+        Result r;
+        if (x.limb_count() >= y.limb_count()) {
+            copy_value(r, x, Result::carry_headroom(x));
+            r.add_in_place(y.representation(), y.is_negative());
+            return r;
+        }
```
eisenwave, line 1157: I think this approach is still pessimistic.

Couldn't you do an addition in an inplace buffer (possibly a new `std::array` for the non-mutating version) and only allocate on carry? If the carry happens, you could copy all the already computed limbs into the new buffer. This would never guess wrong.

The reason I'm concerned is because it may be interesting to have a hard requirement for all operations not to allocate if the result fits into `fixed_buffer`.

include/beman/big_int/big_int.hpp
```diff
@@ -365,15 +365,113 @@ class BEMAN_BIG_INT_TRIVIAL_ABI basic_big_int {
...
+    // Allow extra space to be stated upfront so that functions that know they are
+    // about to grow by a fixed amount (e.g., a carry-out of one
+    // limb in addition) can ask for that space up front.
+    //
+    // Defined inline because the friend template declaration matches the enclosing
+    // class by name, which is tidier than reconstructing the template parameters
+    // out-of-class.
+    template <class Src>
+        requires std::same_as<std::remove_cvref_t<Src>, basic_big_int>
+    friend constexpr void copy_value(basic_big_int& dst, Src&& src, const std::size_t extra_space = 0) {
```
eisenwave: Why does this need to be a `friend`? It seems more natural for this to be an `assign` or `set` member function.

include/beman/big_int/big_int.hpp
```diff
@@ -365,15 +365,113 @@ class BEMAN_BIG_INT_TRIVIAL_ABI basic_big_int {
...
+            if constexpr (is_move) {
+                dst.m_alloc = std::move(src.m_alloc);
+            } else {
+                dst.m_alloc = src.m_alloc;
+            }
+            limb_type* const       dst_limbs = dst.limb_ptr();
+            const limb_type* const src_limbs = src.limb_ptr();
+            for (std::size_t i = 0; i < src_count; ++i) {
+                dst_limbs[i] = src_limbs[i];
+            }
```
```suggestion
            std::copy_n(src_limbs, src_count, dst_limbs);
```
I think for copies specifically, we should try to use `<algorithm>` because that gets manually turned into `memmove` when possible instead of relying on compiler optimizations.

include/beman/big_int/big_int.hpp
mborland: It generally matches the other copy functions. I can convert it to `assign_value` as a member.

eisenwave (approved):

PR #56, opened 2026-04-14
Add [big.int.conv]

ghost:


include/beman/big_int/big_int.hpp
```diff
@@ -948,6 +953,32 @@ constexpr std::strong_ordering operator<=>(const L& lhs, const R& rhs) noexcept
     }
 }
 
+template <std::size_t b, class A>
+template <class T>
+    requires detail::cv_unqualified<T> && std::is_arithmetic_v<T>
```
eisenwave: I think we need our own custom `detail::arithmetic` concept or this won't accept `_BitInt`.

include/beman/big_int/big_int.hpp
```diff
@@ -948,6 +953,32 @@ constexpr std::strong_ordering operator<=>(const L& lhs, const R& rhs) noexcept
...
+constexpr basic_big_int<b, A>::operator T() const noexcept {
+    if constexpr (std::is_same_v<T, bool>) {
+        return !is_zero();
+    } else if constexpr (std::is_floating_point_v<T>) {
+        constexpr T       two64 = static_cast<T>(1ULL << 32) * static_cast<T>(1ULL << 32);
+        T                 result{0};
+        const auto* const limbs = limb_ptr();
+        for (std::size_t i = limb_count(); i-- > 0;) {
+            result = result * two64 + static_cast<T>(limbs[i]);
```
eisenwave, line 999: After https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2026/p3899r1.html is merged into C++, when this hits infinity, that will disqualify `operator float` from being a constant expression.

However, an explicit conversion to a floating-point type should silently produce infinity, like in the following:
```cpp
constexpr auto x = static_cast<float>(static_cast<unsigned _BitInt(256)>(-1));
static_assert(x == __builtin_inf()); // Clang accepts
```

I think you need to detect what the minimum power of two is that is considered infinity, and if `width_mag()` exceeds that value, you silently return infinity.

include/beman/big_int/big_int.hpp
```diff
@@ -948,6 +953,32 @@ constexpr std::strong_ordering operator<=>(const L& lhs, const R& rhs) noexcept
...
+        }
+        return is_negative() ? -result : result;
+    } else {
+        using U = std::make_unsigned_t<T>;
```
eisenwave, line 1012: This should use `inplace_to_bit_uint()` if possible because the entire conversion becomes a compiler intrinsic. That actually applies to the top level now thinking about it, not just the integer case.

tests/beman/big_int/conversion.test.cpp
```diff
@@ -0,0 +1,69 @@
...
+    EXPECT_DOUBLE_EQ(static_cast<double>(x), 1e15);
+    beman::big_int::big_int neg(-42);
+    EXPECT_DOUBLE_EQ(static_cast<double>(neg), -42.0);
+}
+
+TEST(Conversion, ToFloat) {
+    beman::big_int::big_int x(123);
+    EXPECT_FLOAT_EQ(static_cast<float>(x), 123.0f);
+}
+
```
eisenwave, line 95: This needs a test case for converting to infinity.

tests/beman/big_int/conversion.test.cpp
```diff
@@ -0,0 +1,69 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-License-Identifier: BSL-1.0
+
+#include <gtest/gtest.h>
+
+#include <beman/big_int/big_int.hpp>
+
```
eisenwave, line 8: `constexpr` tests needed. That could actually just be a bunch of `static_assert`s.

eisenwave (changes requested):

include/beman/big_int/big_int.hpp
eisenwave: We can do that in a separate PR though.

alcxpr: Will add it in this PR since I'll need it to tests some `static_assert`s.

include/beman/big_int/big_int.hpp
alcxpr: Currently, `std::bit_cast` seems to not support `_BitInt` so I'll guard them with `if BEMAN_BIG_INT_IS_NOT_CONSTEVAL`.

include/beman/big_int/big_int.hpp
```diff
@@ -387,6 +392,19 @@ class BEMAN_BIG_INT_TRIVIAL_ABI basic_big_int {
...
 #endif // BEMAN_BIG_INT_HAS_BITINT
+
+    [[nodiscard]] constexpr auto inplace_to_sbit_int() const noexcept
+#ifdef BEMAN_BIG_INT_HAS_BITINT
+        requires has_inplace_to_bit_uint
+    {
+        static_assert(std::has_unique_object_representations_v<decltype(m_storage.limbs)>,
+                      "Bit-casting doesn't work when there is padding.");
+        const auto mag = std::bit_cast<_BitInt(inplace_bits)>(m_storage.limbs);
+        return is_negative() ? -mag : mag;
```
eisenwave: This can have undefined behavior, such as when the `big_int` currently represents `ULLONG_MAX`. This cannot be negated, and actually, if the uppermost bit is set (unless all subsequent bits are zero), you cannot form a signed `_Bitint` of equal value.

To make this work, I think you need `_BitInt(inplace_bits + 1)`, or you need to check whether the value actually fits into `_BitInt(inplace_bits)` beforehand.

include/beman/big_int/big_int.hpp
```diff
@@ -948,6 +966,52 @@ constexpr std::strong_ordering operator<=>(const L& lhs, const R& rhs) noexcept
...
+#ifdef BEMAN_BIG_INT_HAS_BITINT
+        if BEMAN_BIG_INT_IS_NOT_CONSTEVAL {
+            if (is_storage_static()) {
+                return static_cast<T>(inplace_to_sbit_int());
+            }
+        }
+#endif
+        using U = std::make_unsigned_t<T>;
+        U                 mag{0};
+        constexpr auto    n     = (sizeof(U) + sizeof(limb_type) - 1) / sizeof(limb_type);
```
```suggestion
        constexpr auto    n     = detail::div_to_pos_inf(sizeof(U), sizeof(limb_type));
```

include/beman/big_int/config.hpp
```diff
@@ -244,6 +244,12 @@ using make_signed_or_unsigned_t =
 template <class T>
 concept cv_unqualified_floating_point = cv_unqualified<T> && std::floating_point<T>;
 
+// Modeled if `T` is an arithmetic type - that is, a signed or unsigned integer type
+// including `_BitInt` or a floating-point type.
+// This extends `std::is_arithmetic_v to cover `_BitInt` types which are not standard integral.
+template <class T>
+concept arithmetic = cv_unqualified<T> && (integral<T> || std::floating_point<T>);
```
eisenwave:
```suggestion
concept cv_unqualified_arithmetic = cv_unqualified<T> && (integral<T> || std::floating_point<T>);
```
"Arithmetic type" includes cv-qualified types, so either this should include cv-qualified types or behave like the core language construct (plus `_BitInt` support).

tests/beman/big_int/conversion.test.cpp
```diff
@@ -0,0 +1,138 @@
...
+static_assert(static_cast<unsigned int>(beman::big_int::big_int{42}) == 42U);
+static_assert(static_cast<unsigned int>(beman::big_int::big_int{0}) == 0U);
+static_assert(static_cast<unsigned int>(beman::big_int::big_int{-1}) == std::numeric_limits<unsigned int>::max());
+static_assert(static_cast<unsigned long long>(beman::big_int::big_int{-1}) ==
+              std::numeric_limits<unsigned long long>::max());
+
+static_assert(static_cast<int>(beman::big_int::big_int{0x1'0000'0042LL}) == 0x42);
+
+// TODO(alcxpr): Float static_assert tests deferred until P3899R1 lands.
```
eisenwave: There is no need to defer here. We already know what `static_cast` does; P3899R1 does not change the behavior of that.

P3899R1 just makes it unambiguous that overflow in an arithmetic expression (like multiplication) is not a constant expression, where there is currently compiler divergence.

However, before and after the paper, we know the behavior should be:
```suggestion
static_assert(static_cast<float>(beman::big_int::big_int(std::numeric_limits<float>::max()) << 1), std::numeric_limits<float>::infinity());
```

eisenwave (changes requested):

alcxpr:
Should be OK now. MSVC doesn't like the `static_assert`s with IILEs so I have conditionally excluded them out.

include/beman/big_int/big_int.hpp
```diff
@@ -399,7 +399,10 @@ class BEMAN_BIG_INT_TRIVIAL_ABI basic_big_int {
     {
         static_assert(std::has_unique_object_representations_v<decltype(m_storage.limbs)>,
                       "Bit-casting doesn't work when there is padding.");
-        const auto mag = std::bit_cast<_BitInt(inplace_bits)>(m_storage.limbs);
+        // Use `inplace_bits + 1` to avoid signed overflow when negating a value
```
eisenwave, line 403: This PR is almost good now, but there needs to be a separate:
```cpp
static constexpr bool        has_inplace_to_bit_sint = inplace_bits < BEMAN_BIG_INT_BITINT_MAXWIDTH;
```
You may be requiring `_BitInt(65)` here, but everything else has only tested for `_BitInt(64)`.

eisenwave (changes requested):

eisenwave (approved):

PR #58, opened 2026-04-14
Add classification utility for binary operators

eisenwave:


mborland (approved):
I like this. It looks more readable than my nested conditionals on what should be occurring. Thanks.

PR #60, opened 2026-04-14
Get rid of the last remaining C++23 artifact

### LINKED ISSUE #59: Ensure C++20 compilation
We should have at least one C++20 workflow to make sure no C++23 features are mandated.

While we compile with C++23, all its uses should be hidden behind feature tests so that they don't become mandatory. Without a workflow, we don't actually know if that goal was achieved.

ckormanyos:
This PR removes the final remaining C++23 artifact and walks the lib back to C++20. The cleverly-defined standard-specific macro (`BEMAN_WHATEVER`) is now used instead of `if !consteval (whatever)`.

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -56,7 +56,7 @@ jobs:
             },
             { "versions": ["14"],
               "tests": [
-                { "cxxversions": ["c++26", "c++23"],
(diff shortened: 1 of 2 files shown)


ckormanyos:
> This PR removes the final remaining C++23 artifact and walks the lib back to C++20.

And thanks to @mborland and advice from @eisenwave also adds some C++20 tests to CI.

PR #62, opened 2026-04-14
Implement operators << and >>

### LINKED ISSUE #55: Implement `operator<<` and `operator>>` using storage optimizations from #54

mborland:
- Implements `operator>>` and `operator<<` by using the existing `shift_left` and `shift_right` functions
- Changes `shift_left` to use CLZ to figure out exactly where the high bit is rather than always allocating an extra limb
- Changes the definition of `max_size()` and `shift_max()` based on conversation on WhatsApp. These now fit inside a `size_t` instead of always being 64-bits.

Closes: #55

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -350,6 +350,7 @@ class BEMAN_BIG_INT_TRIVIAL_ABI basic_big_int {
     // Our max shift is then the number of bits represented in these blocks plus the theoretical 63 or 31
     // that are in the same limb.
     // E.g., 000...01 can be shifted 63 bits to the left before it moves to the next limb
+    // TODO : We discussed reducing the size of the encoding for 32-bit platforms so that this result fits in unsigned long (becomes the same type as the limbs)
     using shift_type                      = unsigned long long;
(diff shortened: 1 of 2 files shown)


include/beman/big_int/big_int.hpp
```diff
@@ -1293,6 +1317,73 @@ constexpr detail::common_big_int_type<L, R> operator-(L&& x, R&& y) {
...
+    } else if constexpr (form == detail::binary_op_form::copy_int) {
+        // lvalue: copy then shift. No extra headroom needed since right-shift
+        // only shrinks. The result keeps the source's heap buffer (if any)
+        // even if the shifted value would fit inline.
+        Result r;
+        r.assign_value(x);
+        r.shift_right(shift);
+        return r;
+    }
+}
```
eisenwave, line 1385: We can fix that in a separate issue, but this part is pessimistic. If I compute whatever number shifted to the right by a billion bits or something, then `r.assign_value(x)` pessimizes by copying the original limbs despite those limbs being all discarded by the huge shift anyway.

I think you really need a separate copying form of the `>>` operator this time.

eisenwave (approved):

include/beman/big_int/big_int.hpp
mborland: Opened an issue, and will address.

PR #67, opened 2026-04-15
Optimize Right Shifting

### LINKED ISSUE #65: Right Shift Optimization
Image

mborland:
Right shifting now has three cases:
1) Total loss of information. Return either -1 or 0 depending on the signbit
2) Partial loss of information. Make a copy of only the limbs we need and then make a partial shift. In the negative value case we then need to make sure that we are rounded correctly.
3) No loss of information. Make a full copy and shift accordingly

Closes: #65

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -1479,36 +1479,58 @@ constexpr std::remove_cvref_t<T> operator>>(T&& x, const S s) {
         // For negative values we need to account for the rounding at the end
         //
         // If neither of these situations applies, make a full copy and shift.
+
+        using limb_type = uint_multiprecision_t;
+
         const shift_type shifted_limbs = shift / Result::bits_per_limb;
         const shift_type shifted_bits  = shift % Result::bits_per_limb;
+        const shift_type x_width_mag   = static_cast<shift_type>(x.width_mag());
 
         // Case 1: Everything is discarded except the sign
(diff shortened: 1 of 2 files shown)


include/beman/big_int/big_int.hpp
```diff
@@ -1468,9 +1468,69 @@ constexpr std::remove_cvref_t<T> operator>>(T&& x, const S s) {
...
+        // Case 1: Everything is discarded except the sign
+        if (shift > static_cast<shift_type>(x.width_mag())) {
+            if (x.is_negative()) {
+                return Result{-1};
+            }
+            return Result{};
+        }
+
+        // Case 2: Some of the limbs are discarded
+        if (shifted_limbs != 0) {
```
eisenwave: This condition is not aggressive enough. It's possible that I do `(1n << 64) >> 1`, where I start out with $2^{64}$ before the shift (requiring two limbs), and right-shifting by `1` would keep `shifted_limbs` at zero.

Despite that, the result is $2^{63}$ which only requires a single limb and which requires no allocation to be made, so no allocation should be made in that scenario.

eisenwave (changes requested):

include/beman/big_int/big_int.hpp
eisenwave: And if the condition is not satisifed, you fall back into the naive case where you copy and shift, but that will make a copy of two limbs initially, causing allocation.

eisenwave (approved):

PR #70, opened 2026-04-16
Implement `operator*` and `operator*=`

### LINKED ISSUE #68: Implement `operator*`
In [boost.multiprecision](https://github.com/boostorg/multiprecision/blob/develop/include/boost/multiprecision/cpp_int/multiply.hpp) we have a complete and usable version of Karatsuba for multiplication. There's a few of the helper functions that we don't need anymore, but for the most part we should be able to transpose this to our new representation. This impl uses pre-allocated storage along with the LHS and the RHS. Using our same enum logic, at least in the copy cases we can request all the extra limbs up front in order to do this inplace. Depending on the result size of LHS * RHS there may not really be any advantage to stealing storage that we end up reallocating for anyway. As always correctness comes before optimization.

CC: @ckormanyos

mborland:
- Implements the multiplication routines from `cpp_int`
- We use textbook long multiplication under 40 limbs which is directly from Boost
- We use Karatsuba for large values multiplication
- Like Boost we have a giant scratch pad that we use that keeps us down to one allocation for the scratch pad. We use a LIFO bump-allocator so each "deallocate" as we recurse is moving a pointer rather than a true de-allocation. 
- Slightly re-organizes by moving these detail headers into a detail folder  
- Karatsuba tests are auto generated. Uses python big int and then dumps the limbs into the cpp file so they can be imported. We can probably add fuzzing against cpp_int to check the bytes inside the representation.

Closes: #68


include/beman/big_int/detail/mul_impl.hpp
```diff
@@ -0,0 +1,402 @@
...
+    constexpr void deallocate(const std::size_t n) noexcept {
+        BEMAN_BIG_INT_DEBUG_ASSERT(n <= m_offset);
+        m_offset -= n;
+    }
+};
+
+BEMAN_BIG_INT_DIAGNOSTIC_POP()
+
+// Trim leading zero limbs returning the effective size
+constexpr std::size_t trim_size(const std::span<const uint_multiprecision_t> s) noexcept {
```
eisenwave: This name really has me stumped because it sounds like an operation that modifies `s`.

It should be something like `trimmed_size` or `canonical_size` to make it clear it's a non-mutating operation.

include/beman/big_int/big_int.hpp
```diff
@@ -290,6 +291,9 @@ class BEMAN_BIG_INT_TRIVIAL_ABI basic_big_int {
     template <detail::signed_or_unsigned S>
     constexpr basic_big_int& operator<<=(S s);
 
+    template <detail::arbitrary_integer T>
+    constexpr basic_big_int& operator*=(const T& rhs);
```
```suggestion
    template <detail::common_big_int_type_with<basic_big_int> T>
    constexpr basic_big_int& operator*=(const T& rhs);
```
It used to be `arbitrary_integer` in an older draft, but this would allow operations between `basic_big_int` with different allocators, and that may just be too permissive.

include/beman/big_int/big_int.hpp
```diff
@@ -1795,6 +1808,88 @@ constexpr void basic_big_int<b, A>::add_in_place(const std::span<const uint_mult
...
+
+    // Zero * anything = positive 0
+    if (detail::is_span_zero(a_trimmed) || detail::is_span_zero(b_trimmed)) {
+        limb_ptr()[0] = 0;
+        set_limb_count(1);
+        set_sign(false);
+        return;
+    }
+
+    const std::size_t result_size = a_trimmed.size() + b_trimmed.size();
```
eisenwave, line 1830: This is definitely hugely pessimistic, but we can be faster later.

I think even in the most trivial cases like `2 * 2`, there is at least one limb on each side, there are two limbs allegedly needed here, and you grow to two limbs. Other than multiplication with zero, it seems like allocation always happens.

This PR is already big enough, so it's better to tackle that separately.

eisenwave (changes requested):
Amazing job, looks mostly good. The optimizations can be done later, just gotta get the constraints right.

eisenwave (approved):

PR #72, opened 2026-04-17
Add bitwise operators

### LINKED ISSUE #64: Implement `operator&`, `operator|`, and `operator^`
These all have eight cases similar to `operator-` and `operator+`, but I think all the bitwise operations can be handled in bulk, using a function template operating on the magnitude, that has template parameters
- `enum struct bitwise_op` (AND, OR, XOR)
- `bool neg_left`
- `bool neg_right`

In the function template, you can then use `if constexpr` to perform two's complement emulation for either side using additional bitwise complement and add-with-carry on the magnitude. There also needs to be conversion from the two's complement result back to the magnitude if the result is negative, which can be computed from `neg_left` and `neg_right`.

I think doing it as a template like this is really natural because you can put everything into one loop and not worry about excessive branching within the loop.

ghost:
This closes #64.


include/beman/big_int/big_int.hpp
```diff
@@ -1425,6 +1445,66 @@ constexpr detail::common_big_int_type<L, R> operator-(L&& x, R&& y) {
...
+            x.representation(), x.is_negative(), detail::to_fixed_span(y_limbs), detail::integer_signbit(y));
+    } else {
+        const auto x_limbs = detail::to_limbs(detail::uabs(x));
+        return Result::template dispatch_bitwise<detail::bitwise_op::and_>(
+            detail::to_fixed_span(x_limbs), detail::integer_signbit(x), y.representation(), y.is_negative());
+    }
+}
+
+template <class L, class R>
+constexpr detail::common_big_int_type<L, R> operator|(L&& x, R&& y) {
```
eisenwave: I really don't like how much logic is repeated for `operator|` and `operator^` that is completely identical to `operator&`. All these operators should immediately forward to some common member function with a `detail::bitwise_op` constant template parameter.

include/beman/big_int/big_int.hpp
```diff
@@ -1843,6 +1923,115 @@ constexpr void basic_big_int<b, A>::multiply_into(const std::span<const uint_mul
...
+            }
+        }
+        dst[i] = res;
+    }
+    // If `carry_o` survived all n limbs, we need a final limb of value 1.
+    // This happens e.g. for -1 & -1 = -1 when both inputs have equal size.
+    if constexpr (res_neg) {
+        if (carry_o) {
+            dst[n] = limb_type{1};
+            result.set_limb_count(static_cast<std::uint32_t>(n + 1));
```
eisenwave, line 1977: This seems kinda wrong to me. For `-1 & -1`, isn't `n` equal to `1` (as determined by `std::max`), so you're setting the limb count to `2`, even though only one is needed?

include/beman/big_int/big_int.hpp
```diff
@@ -1843,6 +1923,115 @@ constexpr void basic_big_int<b, A>::multiply_into(const std::span<const uint_mul
...
+            res = l | r;
+        } else {
+            res = l ^ r;
+        }
+        if constexpr (res_neg) {
+            res = ~res;
+            if (carry_o) {
+                ++res;
+                carry_o = res == 0;
+            }
```
```cpp
auto [sum, carry] = detail::carrying_add(res, 0, carry_o);
res = sum
carry_o = carry;
```
The same transformation to use `carrying_add` should be used for the previous operations.

tests/beman/big_int/bitwise.test.cpp
```diff
@@ -0,0 +1,1697 @@
...
+    return result;
+}
+
+template <std::size_t N>
+big_int neg_from_limbs(const std::array<uint_multiprecision_t, N>& limbs) {
+    return -pos_from_limbs(limbs);
+}
+
+} // namespace
+
```
eisenwave, line 35: Correct me if I'm wrong, but what seems to be missing are some of the critical edge cases, like `0 & -1`, `-1 & -1` and such combinations.

eisenwave (changes requested):

include/beman/big_int/big_int.hpp
eisenwave: I think the comment above is just not illustrating where this happens well; it certainly doesn't happen for `-1 & -1` because the `carry_o` refers to a carry resulting from the final negation, but the final negation in that case turns two's complement `-1` (where all bits are set) into `0` by flipping all the bits, then into `1` by incrementing, so there is no carry.

A case where it does happen I think is $(2^{64}-1)$ `xor` $-1$. You have two single-limb operands where after converting two's complement, all bits are set (for different reasons), and XOR between them produces `0`. Since the result sign bit is negative, negation takes place, which sets `carry_o` to `1`.

In this case, the result is mathematically $-2^64$, so you actually do need `set_limb_count(2)` there. Anyway, I just take issue with the comment explaining this in terms of `-1 & -1`, even though the code may be correct.

alcxpr:
Sorry I am a bit late.

alcxpr:
Weird that `git clang-format` doesn't format C++ files.

eisenwave (approved):

PR #76, opened 2026-04-17
Disable MaxSan options for GCC

eisenwave:
GCC sanitizers are too fragile to be used; see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71962

Even something as benign as
```cpp
int main()
{
  static constexpr int x = 0;
  static_assert(bool(&x), "");
}
```
... doesn't work with `-fsanitize=undefined` enabled.


mborland (approved):
Looks good. Hopefully it cycles green, but my local apple clang was fine.

PR #78, opened 2026-04-18
Implement `from_chars`

### LINKED ISSUE #9: Implement `from_chars`
A prerequisite is having multiplication, shifting, and addition. A sample implementation that only needs minor changes can be found here:
https://github.com/eisenwave/cowel/blob/d25b08b1b0e95b4dd981f7589bf3623371ece03b/src/main/cpp/big_int_boost.cpp#L679-L759

Except that there should be no special casing for base 10.

The implementation is also relatively naive, but it's a decent starting point.

### LINKED ISSUE #75: Fully implement `operator""n`

eisenwave:
Fixes #9

This also includes a couple of drive-by changes that were necessary to get the code to run, such as:
- Not using `__builtin_addc` during constant evaluation because it is not `constexpr` on older versions of Clang.
- Fixing some general issues with `carrying_add` and `borrowing_sub`.
- Adding a new constructor that takes iterators, in addition to the `std::from_range` constructor. This is necessary because the `std::from_range` constructor compiles conditionally.

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -2559,11 +2559,11 @@ struct parse_non_allocating {
                 .ec         = std::errc::result_out_of_range,
             };
         }
-        return parse_non_allocating_result{
-            .value      = std::move(parsed),
-            .limb_count = parsed.size(),
-            .ec         = std::errc{},
-        };
+        parse_non_allocating_result r;
+        r.value      = std::move(parsed);
+        r.limb_count = parsed.size();
+        r.ec         = std::errc{};
+        return r;
     }
 
     static constexpr auto result = parse();
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -2400,10 +2400,12 @@ from_chars(const char* const begin, const char* const end, basic_big_int<b, A>& 
     }
 
     // The number of bits per base digit, or zero if the base is not a power of two.
+    // clang-format off
     const auto bits_per_digit = std::has_single_bit(static_cast<unsigned char>(base))
-                                    ? static_cast<unsigned char>(std::countr_zero(static_cast<unsigned char>(base)))
-                                    : static_cast<unsigned char>(0);
-    bool       at_least_one_digit = false;
+                              ? static_cast<unsigned char>(std::countr_zero(static_cast<unsigned char>(base)))
+                              : static_cast<unsigned char>(0);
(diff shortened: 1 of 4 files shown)


eisenwave:
Another day, another compiler bug https://developercommunity.microsoft.com/t/Nonsensical-error-C2440-when-initializin/11077170

include/beman/big_int/big_int.hpp
```diff
@@ -2354,63 +2374,220 @@ using big_int = basic_big_int<64, std::allocator<uint_multiprecision_t>>;
 // [big.int.literal]
 namespace detail {
 
-// Like `std::from_chars`, but detects the base automatically
-// based on the `0x`, `0b`, or `0` prefix.
-template <unsigned_integer T>
+[[nodiscard]] constexpr int digit_value(char c) noexcept {
+    static_assert('A' == 0x41 && 'Z' == 0x5a, "This function requires the ordinary literal encoding to be ASCII.");
+    return '0' <= c && c <= '9'   ? c - '0'
```
mborland, line 2379: An easy optimization here would be to do a single table lookup. It's not a big table.

include/beman/big_int/big_int.hpp
```diff
@@ -2354,63 +2374,220 @@ using big_int = basic_big_int<64, std::allocator<uint_multiprecision_t>>;
...
+            // the output parameter must remain unmodified.
+            out                = {};
+            at_least_one_digit = true;
+        }
+        // In general, each time we concatenate a digit, we multiply with the base and add the digit.
+        // For powers of two, this is equivalent to a left shift, followed by a bitwise OR.
+        // Casting to unsigned char selects a potentially simpler specialization
+        // of the binary operators than using int.
+        if (bits_per_digit) {
+            out <<= bits_per_digit;
```
mborland, line 2426: It might be worthwhile to pre compute how many limbs a valid string of length beginning to end would be and then calling grow up front. You could theoretically have multiple allocations by deferring to the operators since they don't know the theoretical maximum from a complete and valid string.

eisenwave: - https://github.com/eisenwave/std-big-int/issues/84

include/beman/big_int/big_int.hpp
eisenwave: I don't think this matters once parsing is done in blocks of limbs anyway (the utility will be deleted); see
- https://github.com/eisenwave/std-big-int/issues/82

PR #86, opened 2026-04-18
Optimize `from_chars` for powers of two

### LINKED ISSUE #83: `from_chars` should use `|=` instead of `+=` for powers of two, once available
See the TODO.

eisenwave:
Fixes #83

It turns out that we don't need a multiprecision `|=` or `<<=` for powers of two anyway. We can simply append the bits obtained from parsing as many digits as fits into a `uint_multiprecision_t` per iteration.

That is, this PR delegates as much as possible to `std::from_chars`, which is theoretically the fastest way to parse, and could even be turned into an intrinsic in the future.


mborland:
I feel like this could be made significantly simpler if we make this a friend (or a friend `from_chars_impl`) and then add a private index operator to `basic_big_int`. Then you can basically parse a limb size amount of bits via `std::from_chars`, and then write the result directly in. Something like:

```c++
big_int x;
x.grow(bits in valid range from first to last);
size_type current_limb = 0;
While (first != last) // Iterating from end of string to front
{
    Std::from_chars(last - bits per limb, last, x[current_limb], 2);
    // Move last forward limb size bits
    ++current_limb;
    Last -= bits per limb;
}
```

Hopefully this generally makes sense. I think growing during insertion is not ideal, but I believe that's being tracked separately.

eisenwave:
You can only insert this neatly into a specific limb for some powers of two; namely `2` and `16`. This doesn't work for `8` and `32`. There is a TODO, and this should be tracked as a follow-up optimization.

mborland:
> You can only insert this neatly into a specific limb for some powers of two; namely `2` and `16`. This doesn't work for `8` and `32`. There is a TODO, and this should be tracked as a follow-up optimization.

Agreed. More to my `friend` suggestion we could end up something neatly like:

```c++

Namespace detail {

Template <typename big_int>
Friend constexpr from_chars_two_sixteen(char* first, char* last, big_int& x)

Friend constexpr from_chars_ten(...)

}

Template <typename big_int>
[[nodiscard]] from_chars(char* first, char* last, big_int& x, int base)
{
    // Check first > last, sign, etc.

    Switch (base):
        Case 2: from_chars_two_sixteen(first, last, x)
        Case 10: from_chars_ten(first, last, x)
}
```

That would make the overall `from_chars` implementation cleaner without having to make it a friend or require the accessor workarounds.

PR #90, opened 2026-04-18
Optimize `from_chars` for the general case

### LINKED ISSUE #89: `from_chars` should parse blocks of digits
This optimization is already implemented for powers of two, but in the fallback case, the algorithm goes naively digit by digit.

eisenwave:
Fixes #89

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -2576,10 +2576,11 @@ from_chars(const char* const begin, const char* const end, basic_big_int<b, A>& 
     // Parse the valid digit run in blocks that fit into `uint_multiprecision_t`.
     // The first block may be shorter so that all following blocks have the same width,
     // then the accumulated result is built left-to-right by multiplying by `max_pow` before adding each next block.
-    const auto first_block_length =
-        static_cast<std::ptrdiff_t>((current_end - current_begin) % max_digits_per_iteration == 0
-                                        ? max_digits_per_iteration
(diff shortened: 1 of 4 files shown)


include/beman/big_int/big_int.hpp
```diff
@@ -2550,32 +2549,66 @@ from_chars(const char* const begin, const char* const end, basic_big_int<b, A>&
...
+    out = leading_digits;
+
+    for (const char* current_first = current_begin + first_block_length; current_first != current_end;
+         current_first += max_digits_per_iteration) {
+        const char* const current_last = std::min(current_first + max_digits_per_iteration, current_end);
+
+        uint_multiprecision_t        digits         = {};
+        const std::from_chars_result partial_result = std::from_chars(current_first, current_last, digits, base);
+        // Same reasoning as for powers of two; see above.
+        BEMAN_BIG_INT_DEBUG_ASSERT(partial_result.ec == std::errc{});
```
mborland, line 2598: Rather than asserting, shouldn't this be the point where we have to bail out on invalid input? We can't possibly overflow so we only have to check the one case. A failure input would be: 100....0000...a....000.

mborland (commented):
Looks good to me other than the one comment. I appreciate the improved assertion

include/beman/big_int/big_int.hpp
eisenwave: The input was already parsed once up front, which is necessary to determine how many leading digits there are. There can't be invalid input at this point.

PR #94, opened 2026-04-20
Use `_BitInt` for multiplication when possible

eisenwave:
This PR also fixes a bunch of latent bugs that became visible when using `_BitInt` more:
- `width_v` does not accept `_BitInt`
- Hardcoding `if !consteval` to avoid bit-casting to `_BitInt` is not future-proof and does not express intent.
- `inplace_to_bit_uint()` should be conditionally `constexpr`
- `assign_magnitude` has stack buffer overflow if the limb count of `value` is greater than `inplace_capacity`.


mborland (commented):
Good improvement. Thanks

Issue #79, opened 2026-04-18
Memory range defect in Karatsuba multiplication


ckormanyos:
## Describe the deficiency

Sporadic calls of Karatsuba multiplication with mixed limb counts left and right lead to an approximate 1-in-20 allocation problem.

Consider the file `mul_impl.hpp`. In the mul-high path of Karatsuba multiplication, the Debug Assert at [line 221](https://github.com/eisenwave/std-big-int/blob/530112993c4f1d69a488b3531801ed0f93e9da77/include/beman/big_int/detail/mul_impl.hpp#L221) fails on occasion. The call stack reveals it has been called from [line 280](https://github.com/eisenwave/std-big-int/blob/530112993c4f1d69a488b3531801ed0f93e9da77/include/beman/big_int/detail/mul_impl.hpp#L280), which itself has been called from[line 389](https://github.com/eisenwave/std-big-int/blob/530112993c4f1d69a488b3531801ed0f93e9da77/include/beman/big_int/detail/mul_impl.hpp#L389).

## To Reproduce

It is, at the moment, cumbersome to fill limbs in the type. The reproducing program is, therefore, rather large. I can attempt to reduce it soon. The program selects random limb widths ranging from $2{\ldots}128$, These are multiplied. A control check is performed with Boost.Multiprecision's `cpp_int` type.

Running 128 trials always ends up in a crash.

```c++

#include <beman/big_int/big_int.hpp>
#include <boost/multiprecision/cpp_int.hpp>

#include <chrono>
#include <cstdint>
#include <iostream>
#include <random>
#include <utility>

namespace local {

namespace detail {

template<typename IntegralTimePointType,
          typename ClockType = std::chrono::high_resolution_clock>
auto time_point() -> IntegralTimePointType;

template <std::size_t N>
auto make_from_limbs(const std::array<beman::big_int::uint_multiprecision_t, N>& limbs) -> beman::big_int::big_int;

using random_engine_limb_type   = ::std::mt19937_64;
using random_engine_length_type = ::std::linear_congruential_engine<::std::uint32_t, UINT32_C(48271), UINT32_C(0), UINT32_C(2147483647)>;

random_engine_limb_type generator_limb { detail::time_point<typename random_engine_limb_type::result_type>() };
random_engine_length_type generator_length { detail::time_point<typename random_engine_length_type::result_type>() };

[...]

PR #95, opened 2026-04-20
Issue79

ckormanyos:
The purpose of this PR is to handle #79.

Additional modifications include:

- Add Karatsuba edge-case test having at the moment but one test case. More will be added in the near future.
- Compatibility with GCC11, since I would like to use this locally (can be discussed)
- Add more use of the word `typename`.


include/beman/big_int/detail/mul_impl.hpp
```diff
@@ -239,8 +239,8 @@ constexpr void multiply_karatsuba(const std::span<uint_multiprecision_t> result,
     // where B = 2^bits_per_limb
     const auto a_l = a.first(std::min(a.size(), n));
 
-    const uint_multiprecision_t zero = 0;
-    const auto                  a_h  = a.size() > n ? a.subspan(n) : std::span<const uint_multiprecision_t>(&zero, 1);
+    const uint_multiprecision_t zero{0U};
+    const auto                  a_h = a.size() > n ? a.subspan(n) : std::span<const uint_multiprecision_t>(&zero, 1);
```
eisenwave, line 243: This seems kinda unnecessary. Can we not get these stylistic changes mixed up in here?

ckormanyos:
Hi Jan (@eisenwave) the Karatsuba-multiplication fix is green and ready to go. I fumbled around with the linter a bit. But at least now I have a more clear understanding of the style in this project. So that was a good learning experience.

Any further comments or suggestions from your side?

Cc: @mborland

include/beman/big_int/big_int.hpp
```diff
@@ -157,7 +158,10 @@ template <bitwise_op op, cv_unqualified_integral T>
     } else if constexpr (op == bitwise_op::xor_) {
         return static_cast<T>(x ^ y);
     } else {
+#if (!defined(BEMAN_BIG_INT_CLANG) && (defined(BEMAN_BIG_INT_GCC) && (BEMAN_BIG_INT_GCC < 13)))
+#else
         static_assert(false, "Unsupported operation.");
+#endif
```
eisenwave: This seems like a lot of logic that keeps getting repeated. I'd rather use a single macro for this like
```cpp
#define BEMAN_BIG_INT_ILL_FORMED(...) static_assert(false, __VA_ARGS__)
```
which expands to nothing if `static_assert(false)` is not supported yet.

include/beman/big_int/big_int.hpp
```diff
@@ -1235,7 +1240,7 @@ constexpr basic_big_int<b, A>::operator T() const noexcept {
         U                 mag{0};
         constexpr auto    n     = detail::div_to_pos_inf(sizeof(U), sizeof(limb_type));
         const auto* const limbs = limb_ptr();
-        for (std::size_t i = 0; i < std::min(n, static_cast<std::size_t>(limb_count())); ++i) {
+        for (std::size_t i = 0; i < (std::min)(n, static_cast<std::size_t>(limb_count())); ++i) {
```
eisenwave: Can we not do that? I'd rather just `#undef min` and `#undef max` than having to do this everywhere, or `#error` when those macros are defined.

tests/beman/big_int/karatsuba_edge.test.cpp
```diff
@@ -0,0 +1,94 @@
...
+    // 16] b =
+    // FromDigits["ea1b63caac19fde047562759deffb93ddfeccead1bc331f243f976b6a771459e727ef0684ee293f273c25a78cea16e40a6bf92f928503dc7bdc96f2d0f504e9454693142c7722ed04b848f26f16012e3667bf90005868b6b1e3221df5db36719351e79bd39b3eb1a1aa7c6a059f9314013e04e423a6a0a8c7c5c6312eec26d4c48510abe82214c541687d3a370999d9865cd9eddb60d18d593e49c0e648ce5df39b836f7d19fb929b5958c73a63f5d2fcd6de686d29012b021a90afebfc31a7c1dcd44cdeb7b5ed69cda0932cb29960af9e917db13a200668f503f7b11725aed38c2a7186dbf3a47c12a1f23bbc1d0de8f37e99da09d3b3ddfefa5cfa3fcb14dec5810531c6afee53496cc84d7b957b8cb27596b279366f0f25a13a9603bd4d5912f33a69034c359bc850a60f8019bb7342e9dca3f8f1d68fb73da9fc7b83689",
+    // 16] a b IntegerString[%, 16]
+    //
+
+    const std::string str_a(
```
eisenwave, line 20: You don't need to parse from strings manually. This can simply use integer literals like in https://github.com/eisenwave/std-big-int/issues/79#issuecomment-4275856724

eisenwave (changes requested):

include/beman/big_int/big_int.hpp
ckormanyos: I will try this

ckormanyos:
Hi Jan (@eisenwave) I started to preliminarily handle some of the review comments. But I feel like a few more rounds of negotiations might be needed.

include/beman/big_int/detail/config.hpp
```diff
@@ -25,6 +25,18 @@
     #define BEMAN_BIG_INT_HAS_BUILTIN(...) 0
 #endif // __GNUC__
 
+// Undefine min()/max() for MSVC ===============================================
+#if defined(BEMAN_BIG_INT_MSVC)
+    #define NOMINMAX
+#endif
```
eisenwave: Unfortunately this depends on include order.
```suggestion
#ifdef min
    #error min is defined as a macro. Define NOMINMAX.
#endif
#ifdef max
    #error min is defined as a macro. Define NOMINMAX.
#endif
#ifdef BEMAN_BIG_INT_MSVC
    #define NOMINMAX
#endif
```

include/beman/big_int/detail/config.hpp
```diff
@@ -25,6 +25,18 @@
...
+
+// Unsupported static_assert to nothing (for old compilers) ==============
+#if (!defined(BEMAN_BIG_INT_CLANG) && (defined(BEMAN_BIG_INT_GCC) && (BEMAN_BIG_INT_GCC < 13)))
+    #define BEMAN_BIG_INT_STATIC_ASSERT_LEGACY_COMPILERS(...)
+#else
+    #define BEMAN_BIG_INT_STATIC_ASSERT_LEGACY_COMPILERS(...) static_assert(false, __VA_ARGS__)
```
eisenwave: I really don't think this name expresses the intent very well. `BEMAN_BIG_INT_STATIC_ASSERT_FALSE` or `BEMAN_BIG_INT_ILL_FORMED` would be better.

The fact that it's a macro should in itself be evidence that some compiler compatibility stuff is going on, and the name doesn't express that it's `static_assert(false)` specifically.

eisenwave (changes requested):

eisenwave:
> Hi Jan (@eisenwave) I started to preliminarily handle some of the review comments. But I feel like a few more rounds of negotiations might be needed.

Yeah, thanks a lot for fixing this; just gotta iron out the details. Mostly stylistic stuff.

ckormanyos:
OK Jan (@eisenwave), we are coming down to the wire.

- Style seems to be an issue. So if you have any more comments, I'd be happy to address these in this PR.
- Also I can't get huge literals working locally  with `operator""_n`, unfortunately. MSVC claims that the literal is too long and older GCC11 simply does not parse it, leaving no sensible comment other than a listing of the red-colored literal and a caret indicating where pasing stopped. We will have to face huge literals, maybe another day. But this issue WILL creep up. It _might_ even get to core language and the maximum limits of literals (unless this was already done to some larger, sensible value).

See also #96

tests/beman/big_int/karatsuba_edge.test.cpp
ckormanyos: See also #96

ckormanyos: I am resolving this for this PR, since I can't use `operator""_n` on the new tests, at least in this PR at this moment in time. I created #96 to track this phenomenon.

tests/beman/big_int/karatsuba_exercise.test.cpp
```diff
@@ -0,0 +1,202 @@
+
```
mborland:
```suggestion
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
// SPDX-License-Identifier: BSL-1.0
```

tests/beman/big_int/karatsuba_exercise.test.cpp
```diff
@@ -0,0 +1,202 @@
...
+    constexpr unsigned trials{128U};
+
+    bool result_is_ok{true};
+
+    for (unsigned index{0U}; index < trials; ++index) {
+        static_cast<void>(index);
+
+        // TODO(ckormanyos): Having control values would be nice. I wonder if it
+        //                   would be possible to clone Boost.Config and Boost.Math
+        //                   and obtain control values from Boost?
```
mborland, line 192: I can configure the CI in a different PR to fetch content: math, multiprecision, and config

mborland (commented):
This generally looks good to me. Thanks for handling this Chris.

ckormanyos:
> This generally looks good to me.

When you get back, there are a couple optimization and cleanup ideas I have. But nothing critical is pensing at the moment.

PR #98, opened 2026-04-20
Implement `operator/` and `operator%`

### LINKED ISSUE #93: Implement `/` and `%`

mborland:
This is a pretty tight port of `cpp_int`'s division and modulo methods.


include/beman/big_int/detail/div_impl.hpp
```diff
@@ -0,0 +1,332 @@
...
+
+namespace beman::big_int::detail {
+
+// ---------------------------------------------------------------------------
+// Three-way compare of two little-endian unsigned spans.
+// Returns -1 / 0 / +1. Operands need not be trimmed: trailing zero limbs on
+// either side are treated as insignificant.
+// ---------------------------------------------------------------------------
+constexpr int compare_unsigned_spans(const std::span<const uint_multiprecision_t> a,
+                                     const std::span<const uint_multiprecision_t> b) noexcept {
```
eisenwave: Can we stick to using `std::strong_ordering` instead of using `int` for comparisons?

include/beman/big_int/detail/div_impl.hpp
```diff
@@ -0,0 +1,332 @@
...
+        for (std::size_t i = 0; i < shift; ++i) {
+            t_full[i] = 0;
+        }
+
+        uint_wide_t mul_carry = 0;
+        for (std::size_t i = 0; i < divisor.size(); ++i) {
+            mul_carry += static_cast<uint_wide_t>(divisor[i]) * static_cast<uint_wide_t>(guess);
+            t_full[i + shift] = static_cast<uint_multiprecision_t>(mul_carry);
+            mul_carry >>= limb_bits;
+        }
```
eisenwave: It doesn't have to be in this PR, but this should be rewritten in terms of `widening_mul`. `uint_wide_t` is potentially 128-bit, and 128-bit support is generally optional.

include/beman/big_int/big_int.hpp
```diff
@@ -2212,6 +2241,260 @@ constexpr auto basic_big_int<b, A>::operator*=(const T& rhs) -> basic_big_int&
...
+
+template <class L, class R>
+constexpr detail::common_big_int_type<L, R> operator%(L&& x, R&& y) {
+    using Result        = detail::common_big_int_type<L, R>;
+    constexpr auto form = detail::classify_form_v<L, R>;
+
+    Result r;
+    if constexpr (form == detail::binary_op_form::move_move || form == detail::binary_op_form::move_copy ||
+                  form == detail::binary_op_form::copy_move || form == detail::binary_op_form::copy_copy) {
+        r.divmod_into(x.representation(), x.is_negative(), y.representation(), y.is_negative(), false);
```
eisenwave: These call sites are really cryptic with the `false` literal at the end. Can we do something like
```cpp
enum class division_op : bool {
    div, // compute quotient
    rem, // compute remainder
};
```
```suggestion
        r.divmod_into(x.representation(), x.is_negative(), y.representation(), y.is_negative(), detail::division_op::rem);
```

eisenwave (changes requested):

include/beman/big_int/detail/div_impl.hpp
mborland: Fixed in [fbc9d06](https://github.com/eisenwave/std-big-int/pull/98/commits/fbc9d0697b1d9efca7fca659c285a84fb9fdebbd) rather than deferring

eisenwave (approved):

PR #101, opened 2026-04-20
Add testing utilities for big_int vs cpp_int

mborland:
Intention is that you can do:

```c++
EXPECT_TRUE(check_cpp_int_equal(std::plus<>{}, random_big_int(1024), random_big_int(1024)));
```

Or whichever op you want, and it will automatically compare the values in the background. `random_big_int` takes two arguments, number of bits, and sign following signbit convention.

CC: @ckormanyos

### DIFF
--- a/tests/beman/big_int/boost_mp_testing.hpp
+++ b/tests/beman/big_int/boost_mp_testing.hpp
@@ -8,11 +8,18 @@
 
 #include <gtest/gtest.h>
 
-#include <boost/multiprecision/cpp_int.hpp>
-
 #include <beman/big_int/big_int.hpp>
 
+BEMAN_BIG_INT_DIAGNOSTIC_PUSH()
+BEMAN_BIG_INT_DIAGNOSTIC_IGNORED_GCC("-Warray-bounds") // This causes way too many problems.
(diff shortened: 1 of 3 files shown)


ckormanyos:
> Or whichever op you want, and it will automatically compare the values in the background. random_big_int takes two arguments, number of bits, and sign following signbit convention.

Wow, that certainly looks convenient. Thanks for doing this, Matt.

tests/beman/big_int/boost_mp_testing.hpp
```diff
@@ -0,0 +1,180 @@
...
+namespace beman::big_int::boost_mp_testing {
+
+namespace detail {
+
+using cpp_int = ::boost::multiprecision::cpp_int;
+
+// Generates a random hex string for a size of exactly `bits` bits,
+// meaning the MSB is 1 (so the value is in [2^(bits-1), 2^bits)).
+// `bits == 0` returns "0".
+inline std::string random_hex_of_bits(std::mt19937_64& rng, std::size_t bits) {
```
eisenwave: Can we be more consequential about sprinkling `[[nodiscard]]` on results and `const` on parameters where possible?

eisenwave (approved):

ckormanyos:
I am already using it in both the repo as well as locally. Thanks everyone!

Issue #100, opened 2026-04-20
Add testing iterations against `cpp_int`


mborland:

mborland:
Closed by #101 and #103

PR #103, opened 2026-04-20
Modernize Karatsuba_exercise as per 100

ckormanyos:
This simple branch modernizes the Karatsuba-exercise test to use `big_int`-versus`cpp_int` test harness.


ckormanyos:
Use the new test harnesss from #100 and merged in #101

PR #107, opened 2026-04-21
Fix self alias case in `opreator*=`

### LINKED ISSUE #104: Huge multiplication or something along the way fails
## Describe the deficiency

Motivated by Karatsuba multiplication, we test the limits of this brand new type.

Consider a medium-sized Mersenne-prime $2^{1398269}-1{\approx}8.147175644125731{\times}10^{420920}$. We compute this with Boost `cpp_int` correctly and then try with `big_int`.

We use a local definition of the ladder-method to compute the high power.

Boost gets it whereas `big_int` needs some work, it might be a little or a lot.

## To Reproduce

```c++
// Deactivate this to use big_int
#define BEMAN_BIG_INT_EXAMPLE_MERSENNE_USE_BOOST

#if defined BEMAN_BIG_INT_EXAMPLE_MERSENNE_USE_BOOST
#include <boost/multiprecision/cpp_int.hpp>
#else
#include <beman/big_int/big_int.hpp>
#endif

#include <iostream>

namespace local {

template<typename IntegralType>
constexpr auto pow(const IntegralType& b, const unsigned p) -> IntegralType {
    using local_integer_type = IntegralType;

    // Calculate (b ^ p).

    local_integer_type result { };

    result = local_integer_type(static_cast<std::uint8_t>(UINT8_C(1)));

    local_integer_type y       { b };
    local_integer_type p_local { p };

    while(p_local != 0) // NOLINT(altera-id-dependent-backward-branch)
    {
        if(static_cast<unsigned>(p_local & 1U) != 0U)
        {
          result *= y;
        }

        y *= y;

        p_local >>= 1U;
    }

    return result;
}

} // namespace local

auto main() -> int {
#if defined BEMAN_BIG_INT_EXAMPLE_MERSENNE_USE_BOOST
    using mp_int_type = boost::multiprecision::cpp_int;
#else
    using mp_int_type = beman::big_int::big_int;
#endif

    // Mersenne prime 2^1398269 approx. 8.147175644125731*10^420920

    // Mathematica:
    //
    // 2^1398269
    // N[%]

    const mp_int_type two{2};

    const mp_int_type mersenne_big{local::pow(two, 1398269)};

#if defined BEMAN_BIG_INT_EXAMPLE_MERSENNE_USE_BOOST
    std::cout << mersenne_big << std::endl;
#else
    std::cout << mersenne_big.representation().back() << std::endl;
#endif
}
```

## Expected Behavior

Get the exact digits. Boost gets them and gets the right answer. `big_int` fails early on in the ladder method for power. I suspect it might be a simple casting issue, or something that is simply not yet implemented.

The Boost calculation requires a fraction of a second, but printout takes up to 10 seconds.

## Additional Discussions

Cc: @mborland per our offline discussions.

mborland:
Fixes: #104


include/beman/big_int/big_int.hpp
```diff
@@ -2230,6 +2230,15 @@ constexpr auto basic_big_int<b, A>::operator*=(const T& rhs) -> basic_big_int&
         // then multiply into a new *this.
         const basic_big_int temp = std::move(*this);
         *this                    = basic_big_int{};
+
+        // `rhs` may alias `*this` (e.g. `y *= y`).
+        // After the move, `rhs` still references the moved-from destination, so read both operands from `temp`.
+        if constexpr (std::is_same_v<std::remove_cvref_t<T>, basic_big_int>) {
```
eisenwave: This `if constexpr` is reundant because `common_big_int_type` doesn't allow allocator mixing, so you already know that the type is the same in here.

tests/beman/big_int/multiplication.test.cpp
```diff
@@ -271,4 +297,6043 @@ TEST(Multiplication, MultiLimbSquaring) {
     EXPECT_EQ(r, expected);
 }
 
+TEST(Multiplication, Mersenne) {
+#ifndef _MSC_VER
+    // 2^1398269 - 1
+    const std::string control_val{"8147175644125730751426772643891354260153137830850222710321145104846993\
```
eisenwave: Can we convert this to hex in order to make the test faster, or is the multiplication inside `from_chars` part of the test?

eisenwave: In general, I'm really confused about which parts here are part of the test setup and which parts are actually being tested. If this is not part of the test, you could just do a left-shift and decrement.

eisenwave (changes requested):

tests/beman/big_int/multiplication.test.cpp
ckormanyos: I have a rich set of Mersenne prime tests waiting going up to millions of digits without using control strings. So I can either add my test here... Or wait until this thing cycles green, then re-branch and add my tests

eisenwave: If the control string is not part of the test, this should be
```cpp
const big_int control_val = (1_n << 1398269) - 1;
```

ckormanyos: The HEX string is the numeral 1, followed by a whole bunch of f's. The string is an even multiple of 16 in length. It has something like 350k hexadecimal digits.

mborland: I've changed the test to compare to the hex string and to the left shift.

eisenwave: What do we need the hex string for? Do we not trust left-shifting to work? This is a multiplication test, so it should be fine to test in terms of simpler operations.

Adding thousands of lines of integer literal just seems silly to me.

mborland: Counts as a test against left shifting too right? We don't have anything tested out to 350,000 digits so having two control values that we can match against each other can make sense. Now that we have confidence in all ops Chris can easily add larger Mersenne testing.

eisenwave: If you really insist on testing against a string, make it `std::string("1") + std::string(10000, 'f')` or however many `f`s it takes.

Filling 3000 lines of code with the character `f` is just comical.

eisenwave (approved):

tests/beman/big_int/multiplication.test.cpp
mborland: That part is fair. It's now two lines

PR #108, opened 2026-04-21
Make `narrowing_div` more optimizer-friendly

eisenwave:


include/beman/big_int/detail/wide_ops.hpp
```diff
@@ -413,23 +413,32 @@ template <unsigned_integer T>
     const bool quotient_fits = x.high_bits < y;
     if BEMAN_BIG_INT_IS_CONSTEVAL {
         BEMAN_BIG_INT_ASSERT(quotient_fits);
+    } else {
+        BEMAN_BIG_INT_DEBUG_ASSERT(quotient_fits);
     }
```
mborland, line 418: Do we use this kind of check other places? Might be worth rolling into a single macro in a future PR

mborland (approved):
This looks good, and the benchmarks you sent on WhatsApp early showed its the way to go.

include/beman/big_int/detail/wide_ops.hpp
eisenwave: I was having the same thought. We currently don't; maybe it would make sense to turn all the debug asserts into ones that also trigger during constant evaluation.

PR #114, opened 2026-04-22
Detect compound subtraction by self

### LINKED ISSUE #111: Self-aliasing on `operator-=()`
## Describe the deficiency

Self-aliasing on `operator-=()` in other words `self -= self` should result in $0$.

This is a simplification, and perhaps an optimization.

Similar to #104 which handled multiplication.

eisenwave:
Closes #111.

This also comes with a bit of drive-by refactoring: since we are performing a "set to zero" operation in a number of places, it would be better to express that as `set_zero()`.

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -227,6 +227,7 @@ class BEMAN_BIG_INT_TRIVIAL_ABI basic_big_int {
     constexpr void                           set_limb_count(std::uint32_t n) noexcept;
     constexpr void                           set_sign(bool s) noexcept;
     constexpr void                           negate() noexcept;
+    constexpr void                           set_zero() noexcept;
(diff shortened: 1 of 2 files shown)


include/beman/big_int/big_int.hpp
```diff
@@ -730,13 +732,19 @@ constexpr void basic_big_int<b, A>::negate() noexcept {
     }
 }
 
+template <std::size_t b, class A>
+constexpr void basic_big_int<b, A>::set_zero() noexcept {
```
mborland, line 736: Do we also need to set sign here to avoid the negative zero case? We did that at each of the call sites this replaced.

eisenwave: `m_size_and_sign = 1` clears the sign bit.

PR #115, opened 2026-04-22
Refactor constructor from arbitrary-arithmetic

eisenwave:
Some improvements are made here:
- The `BEMAN_BIG_INT_UNSUPPORTED_LONG_DOUBLE` check can be put into `assign_from_float` so it's not repeated.
- The `static_cast` of the argument to `assign_from_float` is unnecessary because lvalue-to-rvalue conversion takes care of that.
- `detail::uabs` can be used instead of manual negation.
- Comparison of `value < 0` is fine; this doesn't need a "correctly typed zero".


mborland (commented):
Good changes. Rare to see a negative diff PR, but they're almost always nice to have.

Issue #123, opened 2026-04-22
Add test for error handling in `from_chars` and `to_chars`


eisenwave:
The tests currently only deal with the happy cases where `from_chars` succeeds and `to_chars` has enough space. To get coverage back up, it will be necessary to write some tests for those.

PR #126, opened 2026-04-22
Improve `to_chars` coverage

eisenwave:
This adds some test for the error handling in `to_chars`.

The `uintmax_t` case is also only instantiated when `uint_multiprecision_t` isn't already as wide, which should also improve coverage data on 64-bit, where `uint_multiprecision_t` and `uintmax_t` are probably the same width anyway.


eisenwave:
See #123

PR #128, opened 2026-04-23
Remove direct usage of `uint_wide_t` from div_impl

### LINKED ISSUE #118: Remove usage of uint_wide_t from div mod implementation

mborland:
Closes: #118

### DIFF
--- a/include/beman/big_int/detail/wide_ops.hpp
+++ b/include/beman/big_int/detail/wide_ops.hpp
@@ -465,7 +465,7 @@ template <unsigned_integer T>
     return {.value = static_cast<T>(result), .borrow = (result >> width_v<T>) != 0};
 }
 
-template <class T>
+template <signed_or_unsigned T>
 struct wide_div_result {
     T       quotient;
     wide<T> remainder;
--- a/include/beman/big_int/detail/wide_ops.hpp
+++ b/include/beman/big_int/detail/wide_ops.hpp
@@ -471,12 +471,6 @@ struct wide_div_result {
     wide<T> remainder;
 };
 
-template <signed_or_unsigned T>
-struct div_result {
-    T quotient;
-    T remainder;
-};
-
 // Returns the quotient and remainder of the division `x / y`.
 // The behavior is undefined if the quotient is not representable as `T`,
 // which is the case if and only if `x.high_bits < y`.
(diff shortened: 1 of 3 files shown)


tests/beman/big_int/wide_ops.test.cpp
```diff
@@ -430,4 +431,82 @@ TEST(WideOps, NarrowingDivUint) {
...
+    EXPECT_EQ(r4.remainder.high_bits, 0ull);
+}
+
+TEST(WideOps, DivideWideByWidePortable) {
+    using beman::big_int::detail::divide_wide_by_wide;
+    using beman::big_int::detail::wide;
+
+    // std::uint16_t never matches uint_multiprecision_t's,
+    // so wider_t<std::uint16_t> is not defined
+    using T = std::uint16_t;
```
eisenwave: This is only a deficiency of `wider_t`. It actually should be written so that it works with any integer type, and then this test case may become dead code. If you specifically want to test the portable implementation, the portable version needs to be a separate function.

eisenwave (changes requested):
Sorry, also needs a rebase now.

tests/beman/big_int/wide_ops.test.cpp
eisenwave: - #144

include/beman/big_int/detail/wide_ops.hpp
```diff
@@ -465,6 +465,18 @@ template <unsigned_integer T>
...
+struct wide_div_result {
+    T       quotient;
+    wide<T> remainder;
+};
+
+template <signed_or_unsigned T>
+struct div_result {
+    T quotient;
+    T remainder;
+};
```
eisenwave: This is duplicating the `div_result` we already have in `config.hpp`.

include/beman/big_int/detail/wide_ops.hpp
```diff
@@ -513,6 +525,60 @@ template <unsigned_integer T>
...
+
+    constexpr std::size_t limb_bits = width_v<T>;
+
+    T q    = 0;
+    T r_lo = 0;
+    T r_hi = 0;
+
+    for (std::size_t i = 2 * limb_bits; i-- > 0;) {
+        const T r_top = static_cast<T>(r_hi >> (limb_bits - 1));
+        r_hi          = static_cast<T>((r_hi << 1) | (r_lo >> (limb_bits - 1)));
```
eisenwave: That seems suspiciously like a `funnel_shl` by one bit.

include/beman/big_int/detail/wide_ops.hpp
```diff
@@ -513,6 +525,60 @@ template <unsigned_integer T>
...
+        const T bit   = (i >= limb_bits) ? static_cast<T>((a.high_bits >> (i - limb_bits)) & T{1})
+                                         : static_cast<T>((a.low_bits >> i) & T{1});
+        r_lo          = static_cast<T>((r_lo << 1) | bit);
+        q             = static_cast<T>(q << 1);
```
eisenwave:
```suggestion
        q <<= 1;
```

eisenwave (changes requested):
Mostly good, but the duplicate `struct` needs to go, and operations should be expressed in terms of our "portable intrinsics" where possible.

eisenwave (approved):

eisenwave:
Actually it would still be nice to use `borrowing_sub` here

PR #133, opened 2026-04-24
Add more tests for consistency with `int` operations

eisenwave:


eisenwave:
Lmao
- add more tests
- coverage decreased by 0.2%

The only way to win is not to play

Issue #122, opened 2026-04-22
Early perf indication add


ckormanyos:
An early performance indication on addition is described.

The code snippet at the bottom shows `big_int`-versus-`cpp_int` in a new Fibonacci calculation. We compute the $1,000,000^{th}$ Fibonacci number.

On Ubuntu, with both types having limb-width $64$, I get:

```bash
stopwatch big_int: 6.37555
stopwatch cpp_int: 3.11337
result_length_is_ok    : 1
result_bytes_same_is_ok: 1
result_is_ok           : 1
```

I will investigate potential bottlenecks in addition.

It is good to catch these early in the game.

```c++

#include <boost/multiprecision/cpp_int.hpp>
#include <beman/big_int/big_int.hpp>

#include <span>

namespace local {

  namespace concurrency {

  struct stopwatch
  {
  public:
    using time_point_type = std::uintmax_t;

    auto reset() -> void
    {
      m_start = now();
    }

    template<typename RepresentationRequestedTimeType>
    static auto elapsed_time(const stopwatch& my_stopwatch) noexcept -> RepresentationRequestedTimeType
    {
      using local_time_type = RepresentationRequestedTimeType;

      return
        local_time_type
        {
            static_cast<local_time_type>(my_stopwatch.elapsed())
          / local_time_type { UINTMAX_C(1000000000) }
        };
    }

  private:
    time_point_type m_start { now() }; // NOLINT(readability-identifier-naming)

    [[nodiscard]] static auto now() -> time_point_type
    {
      timespec ts { };

      const int ntsp { timespec_get(&ts, TIME_UTC) };

      static_cast<void>(ntsp);

      return
        static_cast<time_point_type>
        (
            static_cast<time_point_type>(static_cast<time_point_type>(ts.tv_sec) * UINTMAX_C(1000000000))
          + static_cast<time_point_type>(ts.tv_nsec)
        );
    }

    [[nodiscard]] auto elapsed() const -> time_point_type
    {
      const time_point_type stop { now() };

      const time_point_type
        elapsed_ns
        {
          static_cast<time_point_type>
          (
            stop - m_start
          )
        };

      return elapsed_ns;
    }
  };

  } // namespace concurrency

template <typename BigIntType>
BigIntType fibonacci(unsigned int n) {
    if (n == 0) return BigIntType(0);
    if (n == 1) return BigIntType(1);

    BigIntType a = 0;
    BigIntType b = 1;

[...]

PR #154, opened 2026-04-27
Significant improvements to performance of addition

mborland:
- Combine copy and add
- Manually unroll some of the loops
- Ensure we have at most 1 allocation. In the borrowed storage case this is still deferred to the end

See: #122 and https://github.com/eisenwave/std-big-int/issues/122#issuecomment-4328313614

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -2177,8 +2177,8 @@ constexpr void basic_big_int<b, A>::add_into(const std::span<const uint_multipre
             const alloc_result allocation = alloc_limbs(n);
             limb_type* const   limbs      = allocation.ptr;
 
-            [[maybe_unused]] const bool borrow = run_sub(
-                a, b_span, [limbs](std::size_t idx, limb_type v) { std::construct_at(limbs + idx, v); }, n);
+            [[maybe_unused]] const bool borrow =
+                run_sub(a, b_span, [limbs](std::size_t idx, limb_type v) { std::construct_at(limbs + idx, v); }, n);
             BEMAN_BIG_INT_DEBUG_ASSERT(!borrow);
             for (std::size_t i = n; i < allocation.count; ++i) {
                 std::construct_at(limbs + i);
@@ -2189,9 +2189,9 @@ constexpr void basic_big_int<b, A>::add_into(const std::span<const uint_multipre
             m_storage.data = allocation.ptr;
             finalize_trim_and_sign(limbs, n, a_neg);
         } else {
-            limb_type* const            limbs  = limb_ptr();
-            [[maybe_unused]] const bool borrow = run_sub(
-                a, b_span, [limbs](std::size_t idx, limb_type v) { limbs[idx] = v; }, n);
+            limb_type* const            limbs = limb_ptr();
+            [[maybe_unused]] const bool borrow =
+                run_sub(a, b_span, [limbs](std::size_t idx, limb_type v) { limbs[idx] = v; }, n);
             BEMAN_BIG_INT_DEBUG_ASSERT(!borrow);
             finalize_trim_and_sign(limbs, n, a_neg);
         }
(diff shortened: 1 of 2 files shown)


include/beman/big_int/big_int.hpp
```diff
@@ -2016,6 +2022,210 @@ constexpr void basic_big_int<b, A>::add_in_place(const std::span<const uint_mult
...
+            if (carry) {
+                grow(big + 1);
+                limb_ptr()[big] = limb_type{1};
+                unchecked_set_limb_count(static_cast<std::uint32_t>(big + 1));
+            }
+        }
+
+        unchecked_set_sign(false);
+        if (!is_zero()) {
+            unchecked_set_sign(a_neg);
```
eisenwave: This is a now obsolete idiom that should be replaced with

```suggestion
        unchecked_set_sign(a_neg && !unchecked_is_magnitude_zero());
```
This is better because if `a_neg` is `false`, then short-circuiting kicks in, and checking  whether the magnitude is zero becomes totally unnecessary.

include/beman/big_int/big_int.hpp
```diff
@@ -2016,6 +2022,210 @@ constexpr void basic_big_int<b, A>::add_in_place(const std::span<const uint_mult
...
+        // Sign of the result is `a_neg`.
+        const std::size_t big     = a.size();
+        const std::size_t eff_cap = is_representation_inplace() ? inplace_capacity : m_capacity;
+
+        // `common` is the range where both operands contribute,
+        // beyond it the remaining `a[i]` limbs just propagate the carry.
+        const std::size_t common = b_span.size();
+
+        // Unrolled addition loop
+        auto run_add = [&](auto store) {
```
```suggestion
        const auto run_add = [&](auto store) {
```
Applies to lambdas in general.

eisenwave (changes requested):
This is very promising. I'm not noticing anything major wrong, just some minor stylistic issues.

It would have been interesting to see this performs only with automatic partial unrolling; if the manual unroll doesn't improve things, it should go.

eisenwave:
The diff looks fine, I'm just wondering whether the manual partial loop unroll is doing anything positive in terms of performance, and if so, how much. Do you have a benchmark result where everything is the same in terms of allocations except the partial loop unroll is gone?

I've been burned by trying to be smart with unrolling and SIMD a number of times; modern compilers are clever.

mborland:
I can collect that data point. When I went from the previous implementation to the beginning of `add_into` the performance was still rather bad (around 6.7s to 6.2s).

mborland:
Removing just the manually unrolled loops:

```
stopwatch big_int: 6.52365
stopwatch cpp_int: 5.30032
result_length_is_ok    : 1
result_bytes_same_is_ok: 1
result_is_ok           : 1
Process finished with exit code 0
```

Vs what is in the PR:

```
stopwatch big_int: 3.51187
stopwatch cpp_int: 5.17033
result_length_is_ok    : 1
result_bytes_same_is_ok: 1
result_is_ok           : 1
```

eisenwave (approved):

ckormanyos:
My original Fibonacci test gives very positive results. Thanks Matt. This is much better now. I used my test, not yours and I might have weaknesses in that test. But the results `big_int`-versus-`cpp_int` are essentially the same now. Please go ahead if everything else is OK

```bash
stopwatch big_int: 4.9334
stopwatch cpp_int: 3.09634
result_length_is_ok    : 1
result_bytes_same_is_ok: 1
result_is_ok           : 1
```

mborland:
> My original Fibonacci test gives very positive results. Thanks Matt. This is much better now. I used my test, not yours and I might have weaknesses in that test. But it's all about the same. Please go ahead if everything else is OK
> 
> ```
> stopwatch big_int: 4.9334
> stopwatch cpp_int: 3.09634
> result_length_is_ok    : 1
> result_bytes_same_is_ok: 1
> result_is_ok           : 1
> ```

Could be an architectural or compiler difference? I am using Clang 22.1.4 on ARM64 Mac. You originally ran:

```
stopwatch big_int: 6.37555
stopwatch cpp_int: 3.11337
result_length_is_ok    : 1
result_bytes_same_is_ok: 1
result_is_ok           : 1
```

So getting `big_int` down to 4.93 is still a solid win

ckormanyos:
> Could be an architectural or compiler difference?

Yes exactly. Please proceed full speed ahead. This is a HUGE win and is better than fine now.

ckormanyos:
Even the naive benchmark in my original code is vastly improved.

PR #155, opened 2026-04-27
Add more fast paths and remove inf detection

### LINKED ISSUE #157: `detail::width` does not support `std::_Unsigned128` and `std::_Signed128`

eisenwave:
Single limbs and two limbs should be common cases, so it is worth detecting those and choosing a fast path. In particular, on 32-bit, two two-limb case would correspond to the builtin `uint64_t`-to-floating conversion.

The infinity detection at the end is redundant because the lower-level compose_float function will also produce infinity in the same case.


eisenwave:
I don't even think it's possible to fix this coverage regression on 64-bit because you need a scenario where you have more than two limbs, but still fit both limbs into the mantissa. That's just not going to happen, even with `float128_t`, bit it may happen on 32-bit when converting to `float128_t`.

PR #158, opened 2026-04-27
Add fuzzing of add, sub, mul and punt div to the future

mborland:
Runs the fuzzers for 60s on PR or 300s at 2am UTC


mborland:
Once GitHub works again I'll open an issue to track adding div, but I am working on it now

Issue #121, opened 2026-04-22
Early perf indication mul


ckormanyos:
An early performance indication on multiplication is described.

The code snippet at the bottom shows `big_int`-versus-`cpp_int` in the recent Mersenne calculation.

On Ubuntu, with both types having limb-width $64$, I get:

```bash
stopwatch big_int: 0.0384529
stopwatch cpp_int: 0.0600548
result_length_is_ok    : 1
result_bytes_same_is_ok: 1
result_is_ok           : 1
```

`big_int` seems faster. I believe this advantage comes from multiplication in the Karatsuba range, where we use ultra lightweight `span`s instead of passing references to `cpp_int`. I will investigate further.

```c++

#include <beman/big_int/big_int.hpp>

#include <boost/multiprecision/cpp_int.hpp>

#include <array>
#include <span>

namespace local {

  namespace concurrency {

  struct stopwatch
  {
  public:
    using time_point_type = std::uintmax_t;

    auto reset() -> void
    {
      m_start = now();
    }

    template<typename RepresentationRequestedTimeType>
    static auto elapsed_time(const stopwatch& my_stopwatch) noexcept -> RepresentationRequestedTimeType
    {
      using local_time_type = RepresentationRequestedTimeType;

      return
        local_time_type
        {
            static_cast<local_time_type>(my_stopwatch.elapsed())
          / local_time_type { UINTMAX_C(1000000000) }
        };
    }

  private:
    time_point_type m_start { now() }; // NOLINT(readability-identifier-naming)

    [[nodiscard]] static auto now() -> time_point_type
    {
      timespec ts { };

      const int ntsp { timespec_get(&ts, TIME_UTC) };

      static_cast<void>(ntsp);

      return
        static_cast<time_point_type>
        (
            static_cast<time_point_type>(static_cast<time_point_type>(ts.tv_sec) * UINTMAX_C(1000000000))
          + static_cast<time_point_type>(ts.tv_nsec)
        );
    }

    [[nodiscard]] auto elapsed() const -> time_point_type
    {
      const time_point_type stop { now() };

      const time_point_type
        elapsed_ns
        {
          static_cast<time_point_type>
          (
            stop - m_start
          )
        };

      return elapsed_ns;
    }
  };

  } // namespace concurrency

template <class IntegralType>
constexpr auto pow(const IntegralType& b, unsigned p) -> IntegralType {
    using local_integer_type = IntegralType;

    // Calculate (b ^ p).

[...]

PR #164, opened 2026-04-28
Add multiplication benchmarks

mborland:
See: #121 

To find Mersenne Prime 52 (136279841) we are looking at:

```
stopwatch big_int: 0.168955
stopwatch cpp_int: 0.321642
result_length_is_ok    : 1
result_bytes_same_is_ok: 1
result_is_ok           : 1
```

### DIFF
--- a/tests/beman/big_int/benchmark_testing.hpp
+++ b/tests/beman/big_int/benchmark_testing.hpp
@@ -6,7 +6,7 @@
 
 #include <cstdint>
 #include <ctime>
-
+#include <span>
 #include <beman/big_int/detail/config.hpp>
 
 namespace beman::big_int::benchmark_testing {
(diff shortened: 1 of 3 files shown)


tests/beman/big_int/benchmark_testing.hpp
```diff
@@ -6,7 +6,7 @@
 
 #include <cstdint>
 #include <ctime>
-
```
eisenwave, line 9: This seems accidental. The goal should be to group standard includes.

eisenwave (changes requested):

eisenwave:
Those are some pretty incredible results tbh. No idea why we would be so much faster than Boost.

eisenwave (approved):

eisenwave:
Could ours be faster because of these few lines?
https://github.com/eisenwave/std-big-int/blob/4bfbe363793ca2570b8aabe486bfac79c0d4dec7/include/beman/big_int/big_int.hpp#L2446-L2448

Tbh squaring is not that uncommon of a thing, so maybe it would be worth having a dedicated function for squaring, considering we detect it already.

mborland:
> Could ours be faster because of these few lines?
> 
> https://github.com/eisenwave/std-big-int/blob/4bfbe363793ca2570b8aabe486bfac79c0d4dec7/include/beman/big_int/big_int.hpp#L2446-L2448
> 
> Tbh squaring is not that uncommon of a thing, so maybe it would be worth having a dedicated function for squaring, considering we detect it already.

I don't believe that multiprecision has a special case for squaring. All of the compound operators are in `number` rather than in the specific backends (`cpp_int_backend` in our case) so they do very little: https://github.com/boostorg/multiprecision/blob/965194a9385999964602f01751372f03c8d5ac9a/include/boost/multiprecision/number.hpp#L658

PR #165, opened 2026-04-28
Try support -Wpedantic while having -Werror

ckormanyos:
This pull request attempts to support `-Wpedantic` while having `-Werror`. The separate little file is a trick we utilize in Boost Math/Multiprecision when facing this dilema.

I do not know if folks agree with this way, but I'm throwing it on the table for review, comments, and potential inclusion.

Cc: @eisenwave and @mborland


eisenwave (changes requested):
I don't really see the point in moving things out into a `config128.hpp` file. If we wanted to be consequent about splitting up `config.hpp`, that would involve *many, many* separate headers. A nice side effect of having it all in one is that it's easy to deal with dependencies between features within, by ordering things properly.

I would like to compile with `-Wpedantic` though. It's a useful warning.

eisenwave:
Making `_BitInt` work with `-Wpedantic` will require a `bit_int` alias template that also uses `__extension__` or whatever the thing was to suppress warnings. Same thing for `__int128`.

ckormanyos:
> I would like to compile with `-Wpedantic` though. It's a useful warning.

Well I don't completely approve of this, but it is something.

There is an unmatched `#pragma push whatever`. I have seen this in practice. But this might rule this out as a potential solution.

I am hesitant to decorate 20 places of using `_BitInt` in the code itself for clang. That would be the _last_ _resort_.

ckormanyos:
This one, or a variation thereof is your call Jan (@eisenwave)

Cc: @mborland

ckormanyos:
> I don't really see the point in moving things out into a `config128.hpp` ...

If you want me to retain 1 header, I can provide that variation easily.

It's your choice, Jan.

ckormanyos:
Oh wait, Jan, I forgot. The point of the separate 128 header is to use the pragma for GCC/clang system headers.

The bit-int detection can go back into `config.hpp`.

But the 128 stuff seemingly needs another file.

ckormanyos:
Well this is the draft.

The only thing I do not like is the unmatched and hanging clang diagnostic suppression.

ckormanyos:
I can't push this thing yet.

I don't fully have confidence in it. And I need a second opinion on the hanging `#pragma push whatever`

eisenwave:
> I am hesitant to decorate 20 places of using `_BitInt` in the code itself for clang. That would be the _last_ _resort_.

You don't have to annotate the uses, just the alias template.
```cpp
template<std::size_t N>
using bit_int = _BitInt(N);
template<std::size_t N>
using bit_uint = unsigned _BitInt(N);
```
The alias template doesn't give warnings.

include/beman/big_int/detail/config.hpp
```diff
@@ -110,6 +110,8 @@
     // This case is for Clang when it provides _BitInt as an extension.
     #define BEMAN_BIG_INT_BITINT_MAXWIDTH __BITINT_MAXWIDTH__
     #define BEMAN_BIG_INT_HAS_BITINT 1
+BEMAN_BIG_INT_DIAGNOSTIC_PUSH()
+BEMAN_BIG_INT_DIAGNOSTIC_IGNORED_CLANG("-Wbit-int-extension")
```
eisenwave: You've added a `PUSH` here but I'm not seeing a pop, so it's unsurprising that this blows up.

include/beman/big_int/detail/config128.hpp
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-License-Identifier: BSL-1.0
+
+#ifndef BEMAN_BIG_INT_CONFIG128_HPP
+#define BEMAN_BIG_INT_CONFIG128_HPP
+
+#if defined(BEMAN_BIG_INT_GCC) || defined(BEMAN_BIG_INT_CLANG)
+    #pragma GCC system_header
```
eisenwave: We don't need to make this a system header. It's possible to deal with warnings for `__int128` just fine without that. `__extension__` among other things, and the aliases shouldn't warn, only the use of the `__int128` keyword.

include/beman/big_int/detail/config.hpp
ckormanyos: That was my whole point on this warning. I can not resolve this without decorating many areas in the headers.

I guess we could go for `-Wpedantic` on GCC, but then not on clang. Or we could stop trying for pedantic.

This is not good practice, what I did there. I do not like it

include/beman/big_int/detail/config128.hpp
ckormanyos: I'm not so sure about that. I vaguely remember trying to unify `__float128` with `-Wpedantic` and that is ultimately what we came up with.

include/beman/big_int/detail/config.hpp
ckormanyos: I can't figure out how to disable the bit-extension in a sensible, even manner

ckormanyos:
Ok well this one is tough.

- The word `_BitInt` is used in several code header locations. We would need to eliminate these and use aliases originating from the single point in `config.hpp`.
- The use of `__int128` will ultimately need a separate file. I do not see a way around this and therein use `#pragma GCC system_header`.

I can do all this on another day.

eisenwave:
```cpp
__extension__ typedef signed __int128 int128_t;
__extension__ typedef unsigned __int128 uint128_t;
```

You won't get pedantic warnings if you use the aliases and alias templates everywhere rather than spelling directly.

mborland:
Did pedantic actually emit any warnings besides the extensions? I find that pedantic often is to the detriment of the code base.

eisenwave:
> Did pedantic actually emit any warnings besides the extensions? I find that pedantic often is to the detriment of the code base.

I like `-Wpedantic` because it prevents accidental use of future standards, like getting warnings for use of C++26 extensions available in C++23 mode.

ckormanyos:
OK, this one is now down to the linting only.

This or something like this gives us `-Wpedantic`. There will certainly be some reactions to the methodology here. And I'm flexible on those kinds of coding details. But it proves we can really get `-Wpedantic`.

We can negotiate over names of aliases or if these could/should be in namespaces and/or addition of extra `*.hpp` files (which might be mandatory).

I will now finish linting.

Cc: @eisenwave and @mborland

include/beman/big_int/detail/config.hpp
ckormanyos: A new approach has been taken

include/beman/big_int/detail/config128.hpp
ckormanyos: Done in the new approach

ckormanyos:
> Did pedantic actually emit any warnings besides the extensions? I find that pedantic often is to the detriment of the code base.

I do not want potential library clients (who might use `-Wpedantic`) to be unable to use our newly hatched library.

include/beman/big_int/detail/config.hpp
```diff
@@ -97,19 +97,22 @@
...
 
 #ifdef BITINT_MAXWIDTH
-    // Once _BitInt is a standard feature and available on all compilers,
+    // Once bitint_t is a standard feature and available on all compilers,
     // this case should be selected for all compilers.
     #define BEMAN_BIG_INT_BITINT_MAXWIDTH BITINT_MAXWIDTH
     #define BEMAN_BIG_INT_HAS_BITINT 1
 #elif defined(__BITINT_MAXWIDTH__)
-    // This case is for Clang when it provides _BitInt as an extension.
+    // This case is for Clang when it provides bitint_t as an extension.
```
eisenwave: These comment changes are off. It would be confusing to talk about whether `Clang` provides `bitint_t` as an extension. Please say `_BitInt` here.

include/beman/big_int/detail/config128.hpp
eisenwave: But I still don't see a point in factoring out into a separate header and would still like not to use `system_header` but to suppress more cleanly. See the `__extension__ typedef` approach; that should cover `__int128`, and suppressing Clang warnings should handle `_BitInt`.

include/beman/big_int/detail/config_bitint.hpp
```diff
@@ -0,0 +1,24 @@
...
+#endif
+
+// _BitInt support =====================================================
+
+using bitint32_t   = _BitInt(32);
+using ubitint32_t  = unsigned _BitInt(32);
+using bitint64_t   = _BitInt(64);
+using ubitint64_t  = unsigned _BitInt(64);
+using bitint128_t  = _BitInt(128);
+using ubitint128_t = unsigned _BitInt(128);
```
eisenwave: I don't think we will ever use these. The only one being referenced is `bitint128_t`, but that could just as well use `bitint_n_t<128>`, and that's literally just a single use site. So let's just remove these extra aliases.

include/beman/big_int/detail/config_bitint.hpp
```diff
@@ -0,0 +1,24 @@
...
+template <const unsigned N>
+using bitint_n_t = _BitInt(N);
+template <const unsigned N>
+using ubitint_n_t = unsigned _BitInt(N);
```
eisenwave:
```suggestion
template <std::size_t N>
using bit_int = _BitInt(N);
template <std::size_t N>
using bit_uint = unsigned _BitInt(N);
```
It may seem arbitrary, but these are the exact alias templates proposed in P3666 (though they will be removed next revision), and with good reasons.

eisenwave (changes requested):

include/beman/big_int/detail/config.hpp
ckormanyos: Will do

include/beman/big_int/detail/config_bitint.hpp
ckormanyos: OK

include/beman/big_int/detail/config_bitint.hpp
ckormanyos: Whatever the names are. I just want `-Wpedantic`.

Thank you Jan

ckormanyos: I'll modify all this tomorrow...

ckormanyos: Done. Cycling

include/beman/big_int/detail/config_bitint.hpp
ckormanyos: Done. Cycling

include/beman/big_int/detail/config.hpp
ckormanyos: Done. Cycling

ckormanyos:
Ok I think (or hope) I handled the review. So what's done today is done and maybe this is about right.

We are back to a single config file.
I used standard names.
And I also used `__extension__`

And learned a few tricks along the way...

include/beman/big_int/big_int.hpp
```diff
@@ -1360,8 +1360,8 @@ template <unsigned_integer T>
         // many conditions must be met.
         // For example, `_BitInt(100)` is not eligible due to padding bits,
         // (this could change with a `std::bit_cast_clear_padding` function in the future)
-        // `_BitInt(32)` is too small to be eligible on 64-bit, and
-        // `_BitInt(128)` is eligible, but (as in all other cases),
+        // `_BitInt` is too small to be eligible on 64-bit, and
+        // `_BitInt` is eligible, but (as in all other cases),
```
eisenwave, line 1364: This change seems like an accident.

ckormanyos: Oh I guess I caught these. I hope there's no more to be found.

Nice catch

eisenwave (changes requested):

eisenwave:
I'm terribly sorry but this needs a rebase because one or two more instances of `_BitInt` are now on `main`.

Also that one comment change still needs reverting.

ckormanyos:
> needs a rebase

Done. Good catch. We are close or maybe we got this?

ckormanyos:
Clients can use `-Wpedantic`

ckormanyos:
Hi Jan (@eisenwave), we had a few iterations on this one.

Could you give it a sanity check?

I'll call it done and if I/we missed a comment we can catch up later

eisenwave (approved):

ckormanyos:
Nice.

Thanks everyone for guidance on this seemingly _pedantic_ issue.

PR #168, opened 2026-04-29
Test with multi-limb divisors

eisenwave:
This hopefully increases test coverage after hitting a regression.

I think the reason for the regression is that we don't have tests for multi-limb divisors that go through here.

Also, it seems like we've literally never tested something involving multi-limb integers (e.g. `_BitInt(128)`) so we've never noticed that `uabs` doesn't work with that.


mborland (approved):
Looks good. I'll check the coverage report tomorrow as well to see where else we can improve

PR #169, opened 2026-04-29
Use `divmod_in_place_short` in `to_chars`

eisenwave:


mborland (approved):
3 decimal places for coverage changes seems excessive. Looks good.

PR #170, opened 2026-04-30
Add a copy to runtime function

### LINKED ISSUE #17: Can we circumvent the limitation on constexpr allocation?
It's often useful to have something like:

```cpp
Constexpr std::uint64_t value = function();
```

Can we potentially do something to the affect of:

```cpp
Constexpr auto big_value = copy_to_runtime(big_int_function());
```

This may very well be out of scope for the initial standardization, but I think there is some utility here.

CC: @ckormanyos

mborland:
The `copy_to_runtime` function takes a stateless lambda generator, determines the number of bits that are required to hold the entire value inline, and then creates that statically allocated return value. This allows us to bridge the gap between constant evaluation and runtime. The macro is a much cleaner way to use this functionality, but I recognize that macro functions aren't the most popular thing available.

Closes: #17

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -3828,11 +3828,8 @@ BEMAN_BIG_INT_DIAGNOSTIC_POP()
 
 BEMAN_BIG_INT_DIAGNOSTIC_POP()
 
-// Convenience macros that wrap an expression in a stateless lambda for the user
+// A convenience macro rather than calling a stateless lambda
 #define BEMAN_BIG_INT_COPY_TO_RUNTIME(...) \
-    (::beman::big_int::copy_to_runtime<decltype([]() { return (__VA_ARGS__); })>())
-
-#define BEMAN_BIG_INT_COPY_VALUE_TO_RUNTIME(...) \
     (::beman::big_int::copy_to_runtime<decltype([]() { return (__VA_ARGS__); })>())
 
 #endif // BEMAN_BIG_INT_BIG_INT_HPP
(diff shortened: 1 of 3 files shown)


eisenwave:
I see how it's useful, but I'm not sure how it's useful in this code base. We already have semi-constant initialization via `operator""n`.

Having some kind of `operator""_nx` that returns the right size of `basic_big_int` would be more ergonomic of the goal is just to create constants. The initialization cost is kind of trivial anyway for testing purposes. Anyway, I'm just really confused what this is meant to do if it's not in the paper.

`copy_to_runtime` is also a bizarre name for a `consteval` function.

mborland:
I think that it's going to be a question that people ask early on, since having `constexpr IntegerType integer_constant = some_generator()` is pretty common. Even if it's not in the paper today, we can point people to the answer to this question, be it in the actual code base or just sitting here as a PR.

eisenwave:
> I think that it's going to be a question that people ask early on, since having `constexpr IntegerType integer_constant = some_generator()` is pretty common. Even if it's not in the paper today, we can point people to the answer to this question, be it in the actual code base or just sitting here as a PR.

The problem I see is that this the actual solution to the problem is non-transient allocations, which would require compiler support but provides something much more natural than this thing.

You have the exact same problem with trying to create a `constexpr std::string` or `constexpr std::vector`; this is not a new problem, and I would expect there to be a solution in C++29.

eisenwave:
Feel free to have it in the code base just as a demo, but this is a dead end from a language design standpoint.

mborland:
> > I think that it's going to be a question that people ask early on, since having `constexpr IntegerType integer_constant = some_generator()` is pretty common. Even if it's not in the paper today, we can point people to the answer to this question, be it in the actual code base or just sitting here as a PR.
> 
> The problem I see is that this the actual solution to the problem is non-transient allocations, which would require compiler support but provides something much more natural than this thing.
> 
> You have the exact same problem with trying to create a `constexpr std::string` or `constexpr std::vector`; this is not a new problem, and I would expect there to be a solution in C++29.

I don't disagree that the better solution is to imbue the compiler with additional powers. This works in a not super hackish way today, and theoretically it's another datapoint that non-transient allocations need to be addressed so we don't have to implement these kinds of work arounds in the future.

eisenwave (approved):

PR #174, opened 2026-05-01
Ecc test syntax

ckormanyos:
Improve syntax and robustness of elliptic curve test.

### DIFF
--- a/tests/beman/big_int/ecc_elliptic.test.cpp
+++ b/tests/beman/big_int/ecc_elliptic.test.cpp
@@ -1,20 +1,21 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 // SPDX-License-Identifier: BSL-1.0
 
 // This C++ work has benefited from parts of andreacorbellini/ecc (in Python script).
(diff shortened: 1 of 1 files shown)


tests/beman/big_int/ecc_elliptic.test.cpp
```diff
@@ -62,13 +64,59 @@ using big_sint_backend_type = boost::multiprecision::cpp_int_backend<>;
...
+
+} // namespace local::concurrency
+
 namespace big_int::example {
 
 namespace detail {
 
 auto divmod(const big_sint_type& a, const big_sint_type& b) -> std::pair<big_sint_type, big_sint_type>;
 
 auto divmod(const big_sint_type& a, const big_sint_type& b) -> std::pair<big_sint_type, big_sint_type> {
```
eisenwave, line 113:
```suggestion
auto div_rem_to_neg_inf(const big_sint_type& a, const big_sint_type& b) -> div_result<big_sint_type> {
```
Following the naming convention of https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2026/p3724r3.html

If you follow this convention, things are obvious enough and the comment below isn't needed.

ckormanyos: OK seems good

PR #179, opened 2026-05-02
Implement `|=`, `&=`, and `^=` operators

### LINKED ISSUE #85: Implement `|=`, `&=`, and `^=`
We already have the non-mutating versions of these, so this should be pretty easy.

ghost:
This closes #85.

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -164,6 +164,78 @@ template <bitwise_op op, cv_unqualified_integral T>
     }
 }
 
+// Two's-complement bitwise operation on little-endian unsigned limb spans.
+// Computes `(lhs, lhs_neg) OP (rhs, rhs_neg)` and writes the magnitude of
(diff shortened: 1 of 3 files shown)


include/beman/big_int/big_int.hpp
```diff
@@ -164,6 +164,78 @@ template <bitwise_op op, cv_unqualified_integral T>
...
+            else if constexpr (!neg_left)
+                return lhs.size();
+            else if constexpr (!neg_right)
+                return rhs.size();
+            else
+                return std::max(lhs.size(), rhs.size());
+        } else {
+            return std::max(lhs.size(), rhs.size());
+        }
+    }();
```
eisenwave: This part is copied almost verbatim from `bitwise_in_place`. We should have utility function for determining the limb count from a bitwise operation and limb counts and signs.

eisenwave (changes requested):
Mostly good to go.

eisenwave (approved):

PR #183, opened 2026-05-04
First draft of a motivational example

ckormanyos:
This PR provides a first draft of an introductory motivational example.

There will be several discussions on this, I expect, as we find a common style.


examples/fibonacci.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+auto main() -> int {
+    using big_int_type = beman::big_int::big_int;
+
+    // Compute the 1,000th Fibonacci number.
+    const big_int_type big_int_fibonacci{fibonacci<big_int_type>(1000)};
+
+    constexpr std::size_t fib_buffer_size{256U};
+
+    char* p_str_fib = new char[fib_buffer_size];
```
eisenwave: We have `to_string` support, so manual memory management is really unnecessary here.

examples/fibonacci.cpp
```diff
@@ -0,0 +1,58 @@
...
+        BigIntType next = a + b;
+        a               = b;
+        b               = next;
+    }
+
+    return b;
+}
+
+auto main() -> int {
+    using big_int_type = beman::big_int::big_int;
```
eisenwave:
```suggestion
    using beman::big_int::big_int;
```

eisenwave (changes requested):

examples/fibonacci.cpp
ckormanyos: Done. Thank you

examples/fibonacci.cpp
ckormanyos: Done. It's cleaner. Thx

ckormanyos:
Does it make sense to build all examples in CI?

In some other projects, we do that. Not necessarily execute them, but build them as a syntax, evolutionary check.

PR #185, opened 2026-05-04
Prelim example 512-bit primes via Miller-Rabin

ckormanyos:
$512$-bit pseudo-random primes.

### DIFF
--- a/examples/primes.cpp
+++ b/examples/primes.cpp
@@ -8,7 +8,6 @@
 
 #if defined(BEMAN_BIG_INT_EXAMPLE_PRIMES_USE_ENTROPY)
 auto get_system_entropy() -> unsigned {
-
     std::random_device rd{};
 
     return rd();
@@ -19,7 +18,7 @@ template <class BigIntType, class RndEngineType>
 auto get_pseudo_random_integer(BigIntType& value_to_get, const BigIntType& max_val, RndEngineType& eng) -> BigIntType {
     using distribution_type = std::uniform_int_distribution<std::uint64_t>;
 
-    distribution_type dist{std::uint64_t{UINT64_C(0x1000000000000001)}, std::uint64_t{UINT64_C(0xFFFFFFFFFFFFFFFF)}};
+    distribution_type dist{std::uint64_t{0x1000000000000001}, std::uint64_t{0xFFFFFFFFFFFFFFFF}};
 
     value_to_get = BigIntType{};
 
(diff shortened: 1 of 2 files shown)


ckormanyos:
This is a draft. I need to mark it as a draft.

examples/primes.cpp
```diff
@@ -0,0 +1,198 @@
...
+auto current_time_stamp_now() noexcept -> BuiltInUintType {
+
+    const auto current_now =
+        static_cast<std::uintmax_t>(std::chrono::duration_cast<std::chrono::nanoseconds>(
+                                        std::chrono::high_resolution_clock::now().time_since_epoch())
+                                        .count());
+
+    return static_cast<BuiltInUintType>(current_now);
+}
+#endif // BEMAN_BIG_INT_EXAMPLE_PRIMES_USE_CHRONO_ENTROPY
```
eisenwave: Why not just use `std::random_device` if you need true randomness?

examples/primes.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+template <class BigIntType, class RndEngineType>
+auto get_pseudo_random_integer(BigIntType& value_to_get, const BigIntType& max_val, RndEngineType& eng) -> BigIntType {
+    using distribution_type = std::uniform_int_distribution<std::uint64_t>;
+
+    distribution_type dist{std::uint64_t{UINT64_C(0x1000000000000001)}, std::uint64_t{UINT64_C(0xFFFFFFFFFFFFFFFF)}};
```
eisenwave:
```suggestion
    distribution_type dist{std::uint64_t{0x1000000000000001}, std::uint64_t{0xFFFFFFFFFFFFFFFF}};
```
These examples are user-facing, so I'd avoid these C macros. Also in general, I really don't see a point in them if you wrap the thing in `std::uint64_t{...}` anyway.

examples/primes.cpp
ckormanyos: Sure. True is relative, but that's standard so OK no prob.

examples/primes.cpp
ckormanyos: Sure. OK.

ckormanyos: Missed this one. I'll commit again. Getting close

ckormanyos:
For me, I could do (in time) one advanced example after this. Maybe explore integer roots.
But I think these examples are expressive so far, so that is optional.

I don't want really think it's a good idea to formulate elliptic curve geometry as an example (unless it seems like a good idea to others). It is really complicated and gets to about more than 500 lines.

So if any more examples, I'd do integer roost sqrt, cbrt and rootk.

Thoughts?

PR #198, opened 2026-05-07
Unary ops Boost test harness and fuzz sqrt

ckormanyos:
Add unary ops to Boost teat harness and also add fuzzuing of a locally implementsed `sqrt`.


mborland:
@ckormanyos if you go to the files changed tab on the PR you can click "add to batch" for each linter suggestion and then commit the batch with all changes

tests/beman/big_int/fuzzing/fuzz_common.hpp
```diff
@@ -21,16 +21,19 @@ namespace beman::big_int::fuzz {
 // Layout: top bit of byte 0 = sign; the remaining 7 bits of byte 0 plus
 // bytes 1..n-1 are the magnitude bytes, big-endian. Empty input maps to "0"
 // to avoid producing an invalid (empty) hex literal.
-[[nodiscard]] inline std::string bytes_to_signed_hex(const std::uint8_t* data, const std::size_t size) {
+[[nodiscard]] inline std::string
+bytes_to_signed_hex(const std::uint8_t* data, const std::size_t size, const bool use_negative = true) {
```
mborland, line 25: `use_negative` was added but it doesn't look like we ever actually use that parameter from a call site?

ckormanyos:
I did something silly. I need to do something else for an hour or two then I will fix it. Marked as draft for the moment

mborland:
> I did something silly. I need to do something else for an hour or two then I will fix it. Marked as draft for the moment

Sounds good. I think your fuzz sqrt test just needs a `using beman::big_int::big_int;`

tests/beman/big_int/fuzzing/fuzz_common.hpp
ckormanyos: Somehow I went out to take a bike ride and just reverted an hour's worth of work. I tend to do things like that. Recovery underway... Sigh. Thx Matt

ckormanyos: Done. It looks like I should have done the commit prior to recreation. OK well it's back on track.

ckormanyos:
OK I'm pushing this. This one also reduces the primes run-time scope to a minimal amount

PR #200, opened 2026-05-07
Remove unneeded werror CI runs

mborland:
We already have -Werror applied to every test in every run


eisenwave (approved):

mborland:
The coverage bug is not spurious. I'll open a PR upstream

PR #206, opened 2026-05-08
Make primes example fully robust

ckormanyos:
This PR makes the primes example fully robust. I considered this for a while. But if we publish a primes example, then Miller-Rabin should also handle all edge cases and bbe fully optimized.

### DIFF
--- a/examples/primes.cpp
+++ b/examples/primes.cpp
@@ -109,19 +109,8 @@ auto miller_rabin(const BigIntType& np, const int trials) -> bool {
     // This subroutine returns true if the prime candidate is prime within
     // the limits of Miller-Rabin testing for the given input of trials.
 
-    // Table[Prime[i], {i, 2, 49, 1}] =
-    // {
-    //     3,   5,   7,  11,  13,  17,  19,  23,
-    //    29,  31,  37,  41,  43,  47,  53,  59,
-    //    61,  67,  71,  73,  79,  83,  89,  97,
-    //   101, 103, 107, 109, 113, 127, 131, 137,
-    //   139, 149, 151, 157, 163, 167, 173, 179,
-    //   181, 191, 193, 197, 199, 211, 223, 227
-    // }
-    // See also:
-    // https://www.wolframalpha.com/input/?i=Table%5BPrime%5Bi%5D%2C+%7Bi%2C+2%2C+49%7D%5D
-
-    const std::array<unsigned, static_cast<std::size_t>(UINT8_C(48))> small_primes = {
+    // Table[Prime[i], {i, 2, 49, 1}]
+    constexpr std::array<unsigned, 48> small_primes = {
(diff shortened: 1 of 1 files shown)


examples/primes.cpp
```diff
@@ -109,21 +109,87 @@ auto miller_rabin(const BigIntType& np, const int trials) -> bool {
...
+    //    29,  31,  37,  41,  43,  47,  53,  59,
+    //    61,  67,  71,  73,  79,  83,  89,  97,
+    //   101, 103, 107, 109, 113, 127, 131, 137,
+    //   139, 149, 151, 157, 163, 167, 173, 179,
+    //   181, 191, 193, 197, 199, 211, 223, 227
+    // }
+    // See also:
+    // https://www.wolframalpha.com/input/?i=Table%5BPrime%5Bi%5D%2C+%7Bi%2C+2%2C+49%7D%5D
+
+    const std::array<unsigned, static_cast<std::size_t>(UINT8_C(48))> small_primes = {
```
eisenwave:
```suggestion
    constexpr std::array<unsigned, 48> small_primes = {
```

eisenwave: In general, can we stop using weird `UINT8_C` macros in our examples?

examples/primes.cpp
```diff
@@ -109,21 +109,87 @@ auto miller_rabin(const BigIntType& np, const int trials) -> bool {
...
+            // the lowest limb.
+            bool is_small_prime{false};
+
+            for (const auto& small_p : small_primes) {
+                if (static_cast<unsigned>(n0 == small_p)) {
+                    is_small_prime = true;
+
+                    break;
+                }
+            }
```
eisenwave:
```suggestion
            bool is_small_prime = std::ranges::contains(small_primes, n0);
```

examples/primes.cpp
```diff
@@ -109,21 +109,87 @@ auto miller_rabin(const BigIntType& np, const int trials) -> bool {
...
+            // This test does not include the secondary query if (np == small_p).
+            // This is OK here because exact small primes have already been
+            // filtered out above.
+
+            if ((np % small_p) == static_cast<unsigned>(UINT8_C(0))) {
+                is_small_prime_divisible = true;
+
+                break;
+            }
+        }
```
eisenwave:
```suggestion
        bool is_small_prime_divisible = std::ranges::any_of(small_primes, [](unsigned p) {
            return np % small_p == 0;
        };
```

eisenwave (changes requested):

examples/primes.cpp
ckormanyos: Oh the end of an era. Yeah sure

examples/primes.cpp
ckormanyos: OK

examples/primes.cpp
ckormanyos: OK, nice simplifications.. Thx

ckormanyos:
I did not clean it up as much as wanted. I'll do that now.

PR #210, opened 2026-05-10
Rsa

ckormanyos:
This (probably final) example performs a straightforward $2048$-bit RSA round-trip encryption/decryption.

I'm mostly satisfied, but a few things are awkward and I wonder if there are possible improvements?

Constructing and initializing a `big_int` from a hexadecimal string is carried out with a helper subroutine, but this seems bulky and hard-to-read. I know that string construction will be a controversial topic with lively discussion, but this example would have benefitted from it. The same is true when converting a the `big_int` result to a hexadecimal string, which itself is converted back to ASCII. The latter problem will be solved with future support for `format`, however.


eisenwave (approved):
The uses of `inline` and `static` are weird, especially when combined, but we can fix that later.

ckormanyos:
> The uses of `inline` and `static` are weird, especially when combined, but we can fix that later.

No problem simplifying it now. Thanks Jan

Issue #211, opened 2026-05-11
Is the draft implementation of `size()` in agreement with the draft paper?


ckormanyos:
## Use case

Is `size()` intended for the bit count or for the limb count?

```c++
#include <beman/big_int/big_int.hpp>

#include <iostream>

auto main() -> int {
    using beman::big_int::big_int;

    big_int two_pow_65{big_int{1} << 65};

    // The result is 2, whereby one might expect 65 or 128?
    std::cout << two_pow_65.size() << std::endl;
}
```

## What I like

I appreciate the `size()` method.

## What I dislike

But ithe description of `size()` in the draft paper seemingly disagrees with its draft implementation.

ckormanyos:
I also noticed that `size()` is used in many places in the reference application for internal operations like multiplication. So it can't just be changed in one spot.

eisenwave:
The implementation is severely lagging behind what is proposed in the paper for a few of those functions, and we need to do a pretty large refactor.

ckormanyos:
> The implementation is severely lagging behind what is proposed in the paper for a few of those functions, and we need to do a pretty large refactor.

Sure, no problem. But which value is intended for `size()`? Should it be more like `representation().size()` related to the limb count or should `size()` represent more of the bit count?

Because in the implementaton details, I find `size()` being used frequently to represent the limb count. We would need to change all these if the meaning of `size()` is changed in the implementation.

eisenwave:
We've had a talk about this in our group, and anything that doesn't have `representation` in its name should return the number of `bits`. And yeah, we would need to change all our existing `size()` uses and whatnot.

ckormanyos:
OK. Thank you.

On second glance, I don't think `big_int::size()` it is used as frequently as I had initially suspected. When `size()` appears in the implementation details, it us usually size of a `span<>` being queried.

ckormanyos:
One follow-up question, should `big_int::size()` return a multiple of limb bits? Or shoud the actual bit count of the most significant limb be taken into consideration?

So in the example way at the top of this page, do you think `size()` should return $65$ or $128$?

eisenwave:
It should return 66.

johnmcfarlane:
If this value refers to bits (and not limbs) then the word "size" would conflict with uses of "size" - as well as names for bit counts and bit widths - elsewhere in the standard library and beyond. I'd recommend "digits" or "width". See P0828 for further details.

ckormanyos:
[...]

PR #212, opened 2026-05-11
Attempt to work issue 211 size() method

ckormanyos:
This pull request is intended to handle #211.


eisenwave (changes requested):
I'm pretty sure we already have a function that returns the number in bits. This should all be a simple function rename and not involve any new numerics.

ckormanyos:
> I'm pretty sure we already have a function that returns the number in bits. This should all be a simple function rename and not involve any new numerics.

Um I'm afraid I can't find it, or I overlooked it (the most likely scenario), or it's missing.

You would want to use intrinsics for this one, but that breaks `constexpr`. We could use a `consteval` path for supporting both with/without intrinsics. @mborland has done a lot of work in that area in our other projects.

But I need, first, to see if the `msb`-like function really already exists.

mborland:
> > I'm pretty sure we already have a function that returns the number in bits. This should all be a simple function rename and not involve any new numerics.
> 
> Um I'm afraid I can't find it, or I overlooked it (the most likely scenario), or it's missing.
> 
> You would want to use intrinsics for this one, but that breaks `constexpr`. We could use a `consteval` path for supporting both with/without intrinsics. @mborland has done a lot of work in that area in our other projects.
> 
> But I need, first, to see if the `msb`-like function really already exists.

Do you want me to handle this? `msb` ends up being quite involved since you need platform intrinsics and the fall back to De Bruijn mul. I'm pretty sure I can just copy and paste over well optimized versions from other projects.

mborland:
Actually in C++20, this msb function should just be `std::countl_zero`

ckormanyos:
>> But I need, first, to see if the `msb`-like function really already exists.

> Do you want me to handle this? msb ends up being quite involved since you need platform intrinsics and the fall back to De Bruijn mul. I'm pretty sure I can just copy and paste over well optimized versions from other projects.

Yes, for sure. Nice Matt. Let's break it down.

Down the road, we need consensus on:
- Does the function already exist?
- Can it be optimized (certainly yes in exactly the way you described)?
- What should it be called?
- And should the argument $42$ return $6$ or $7$?

So moving forward, I'd say you just work on my branch (or make another one). The function could initially retain the name `size()`. And down the road it can be called `whatever`.

There is one single test case that needs adaption. You see that in my initial PR attempt.

eisenwave:
> Um I'm afraid I can't find it, or I overlooked it (the most likely scenario), or it's missing.

https://github.com/eisenwave/std-big-int/blob/b8331ecd3f5e27623bc47a6cde124afa8867112a/include/beman/big_int/big_int.hpp#L1229-L1237

ckormanyos:
>> Um I'm afraid I can't find it, or I overlooked it (the most likely scenario), or it's missing.

> Lines 1229 to 1237

Oh wow! Yes, that is it. Sorry I overlooked it and thanks for pointing that out. It already uses the LIB function mentioned by Matt. So this seems done. Let me work that into the function.

Thanks Jan

ckormanyos:
OK, I attempted to use the existing function.

I am a bit uncertain of the separate paths for `is_zero()` and `is_negative()`, and I changed one of the test values.

So please critically review these changes OK Jan?

ckormanyos:
I think this is consistent with the 4444 draft now. It went green.

tests/beman/big_int/allocation.test.cpp
```diff
@@ -10,27 +10,27 @@
...
     beman::big_int::big_int x{
         31415926535897932384626433832795028841971693993751058209749445923078164062862089986280348253421170679821480865132823066470938446095505822317253594081284811174502841027019385211055596446229489549303819644288109756659334461284756482337867831652712019091456485669234603486104543266482133936072602491412737245870066063155881748815209209628292540917153643678925903600113305305488204665213841469519415116094330572703657595919530921861173819326117931051185480744623799627495673518857527248912279381830119491298336733624406566430860213949463952247371907021798609437027705392171762931767523846748184676694051320005681271452635608277857713427577896091736371787214684409012249534301465495853710507922796892589235420199561121290219608640344181598136297747713099605187072113499999983729780499510597317328160963185950244594553469083026425223082533446850352619311881710100031378387528865875332083814206171776691473035982534904287554687311595628638823537875937519577818577805321712268066130019278766111959092164201989_n};
-    return x.size() == 3324; // still inline, size() returns msb
+    return x.size() == 3324; // still inline, size() returns msb + 1
```
mborland: This should not be inline

ckormanyos: Done

ckormanyos:
This PR makes the `size()` method agree with its specification. We added some tests also.

Nobody can at the moment come up with a better name for the `size()` method. So it will remain unchanged until a suitable replacement can be found.

Merging.

ckormanyos:
Fixes #211 (until a better name can be found)

mborland:
@ckormanyos do you remember why you made a copy in the case that the value is negative? see: https://github.com/eisenwave/std-big-int/issues/341

ckormanyos:
Hi @mborland surely I simply overlooked the extra added effort

mborland:
> Hi @mborland surely I simply overlooked the extra added effort

No worries. I wanted to double check and make sure I wasn't missing something that had been previously corrected. Fix is merged.

PR #217, opened 2026-05-12
Implement support for `std::hash`

### LINKED ISSUE #215: Implement support for `std::hash`
Other languages such as python use Siphash 2-4 for this task. Pretty easy to implement, and pretty performant.

mborland:
Applies Siphash 2-4 to the representation for the implementation. The removal of the `BEMAN_BIG_INT_DIAGNOSTIC_POP()` is due to the fact that there were two calls to POP at the bottom of the file. Now it's just one.

Closes: #215

### DIFF
--- a/include/beman/big_int/detail/siphash.hpp
+++ b/include/beman/big_int/detail/siphash.hpp
@@ -83,7 +83,7 @@ constexpr std::size_t siphash(const std::span<const uint_multiprecision_t> limbs
         for (std::size_t i = 0; i < pairs; ++i) {
             auto m = static_cast<std::uint64_t>(limbs[i * 2]) |
                      (static_cast<std::uint64_t>(limbs[i * 2 + 1]) << 32);
-            s = compress(s, m);
+            s      = compress(s, m);
         }
         // Odd trailing uint_multiprecision_t folds into the final block
     }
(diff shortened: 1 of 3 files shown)


mborland:
Need to add the sign in the call to Siphash so that -5 and 5 don't collide

include/beman/big_int/detail/siphash.hpp
```diff
@@ -0,0 +1,111 @@
...
+    auto b = static_cast<std::uint64_t>(byte_len) << 56;
+    if constexpr (sizeof(uint_multiprecision_t) == sizeof(std::uint32_t)) {
+        if (limbs.size() & 1) {
+            b |= static_cast<std::uint64_t>(limbs.back());
+        }
+    }
+
+    compress(v0, v1, v2, v3, b);
+
+    v2 ^= 0xFFULL;
```
eisenwave:
```suggestion
    v2 ^= 0xFEMPTY;
```

include/beman/big_int/detail/siphash.hpp
```diff
@@ -0,0 +1,111 @@
...
+#include <cstdint>
+#include <cstddef>
+#include <bit>
+#include <span>
+
+namespace beman::big_int::detail {
+
+namespace impl {
+
+constexpr void sipround(std::uint64_t& v0, std::uint64_t& v1, std::uint64_t& v2, std::uint64_t& v3) {
```
eisenwave: The codebase generally tries to avoid mutation and uses lots of `const`.

Could we avoid all the mutation and out parameters here if wee just bundled up this state in a `struct`? This code feels very C-style.

eisenwave (approved):

PR #220, opened 2026-05-12
Add and test `pmr::big_int`

### LINKED ISSUE #219: Implement `big_int` with `std::pmr` allocator
From the paper:

```c++
  namespace pmr {
    template<size_t b>
      using basic_big_int =
        std::basic_big_int<b, polymorphic_allocator<uint_multiprecision_t>>;
    using big_int = basic_big_int<std::big_int::inplace_bits>;
  }
```

mborland:
We have to add a few extra cases where the allocator is stateful, which was previously not tested. 

Closes: #219

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -3140,7 +3140,11 @@ template <std::size_t b, class A>
 constexpr auto basic_big_int<b, A>::alloc_limbs(const size_type n) -> alloc_result {
     BEMAN_BIG_INT_ASSERT(n != 0);
 #if defined(__cpp_lib_allocate_at_least) && __cpp_lib_allocate_at_least >= 202302L
-    return alloc_traits::allocate_at_least(m_alloc, n);
+    if constexpr (detail::traits_has_allocate_at_least<alloc_traits, A>) {
+        return alloc_traits::allocate_at_least(m_alloc, n);
+    } else {
+        return {.ptr = alloc_traits::allocate(m_alloc, n), .count = n};
+    }
 #else
     return {.ptr = alloc_traits::allocate(m_alloc, n), .count = n};
 #endif
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -641,28 +641,27 @@ class BEMAN_BIG_INT_TRIVIAL_ABI basic_big_int {
             return;
         }
 
-        constexpr bool    is_move   = !std::is_lvalue_reference_v<Src>;
         const std::size_t src_count = src.limb_count();
         const std::size_t needed    = src_count + extra_space;
         const std::size_t eff_cap   = is_representation_inplace() ? inplace_capacity : m_capacity;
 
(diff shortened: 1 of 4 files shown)


include/beman/big_int/big_int.hpp
```diff
@@ -687,20 +702,46 @@ class BEMAN_BIG_INT_TRIVIAL_ABI basic_big_int {
...
-                src.m_storage.limbs[0] = limb_type{0};
-                if (m_capacity < needed) {
-                    grow(needed);
+                const bool can_steal = []() {
+                    if constexpr (propagate_alloc || alloc_traits::is_always_equal::value) {
+                        return true;
+                    } else {
+                        return false;
+                    }
+                }();
```
```suggestion
                constexpr bool can_steal = propagate_alloc || alloc_traits::is_always_equal::value;
```

include/beman/big_int/big_int.hpp
eisenwave: We can avoid the `is_move` and all the `if constexpr`s checking it using `std::forward_like` I think, or using some utility lambda.

include/beman/big_int/big_int.hpp
```diff
@@ -3206,6 +3251,15 @@ template <typename Generator>
 // Standard public alias for defaulted type
 using big_int = basic_big_int<64, std::allocator<uint_multiprecision_t>>;
 
+namespace pmr {
+
+template <std::size_t b>
+using basic_big_int = beman::big_int::basic_big_int<b, std::pmr::polymorphic_allocator<uint_multiprecision_t>>;
+
+using big_int = basic_big_int<64>;
```
eisenwave: It's fishy that we repeat the `64` magic number here. I would like to guarantee that `big_int` and `pmr::big_int` have the same minimum bits, and the paper does provide that guarantee.

Although thinking about it, the paper maybe shouldn't provide that guarantee because the layout the implementation chooses may be quite different for an empty class allocator like `std::allocator` and for `std::polymorphic_allocator`, which contains a pointer. Nevermind then.

eisenwave (changes requested):

include/beman/big_int/big_int.hpp
mborland: TIL there's an `std::forward_like`.

include/beman/big_int/big_int.hpp
```diff
@@ -640,19 +641,27 @@ class BEMAN_BIG_INT_TRIVIAL_ABI basic_big_int {
...
         if (needed <= eff_cap) {
             // Fast path: current buffer is already big enough
             const auto old_count = limb_count();
             m_size_and_sign      = src.m_size_and_sign;
-            if constexpr (is_move) {
-                m_alloc = std::move(src.m_alloc);
-            } else {
-                m_alloc = src.m_alloc;
+            if constexpr (propagate_alloc) {
+                m_alloc = std::forward_like<Src>(src.m_alloc);
```
```suggestion
                m_alloc = std::forward<Src>(src).m_alloc;
```
Same for all other such uses of `std::forward`.

I guess that would have always been the better refactor, but I didn't notice that it could be written like this.

eisenwave (changes requested):

eisenwave (approved):

PR #223, opened 2026-05-13
Implement Toom-Cook 4

### LINKED ISSUE #214: Implement Toom-Cook 4

mborland:
Similar to Toom-Cook 3 we follow Bodrato (http://marco.bodrato.it/papers/Bodrato2007-OptimalToomCookMultiplicationForBinaryFieldAndIntegers.pdf). Removes some of the lambdas from Toom-Cook 3 and converts them to shared usage free functions. Future work is tuning the memory allocations across the board. After some testing I have 6s here, but Toom-Cook 3 is currently using 8s. We can roll our own allocator with public test probes to figure this out.

Closes: #214 

CC: @ckormanyos

### DIFF
--- a/include/beman/big_int/detail/mul_impl.hpp
+++ b/include/beman/big_int/detail/mul_impl.hpp
@@ -259,6 +259,91 @@ constexpr void multiply_karatsuba(const std::span<uint_multiprecision_t> result,
     scratch.deallocate(total_scratch);
 }
 
+// ---------------------------------------------------------------------------
+// Helpers shared by the Toom-Cook variants. All operate on little-endian
+// multi-precision spans of uint_multiprecision_t limbs.
+// ---------------------------------------------------------------------------
+
+// In-place tmp[0..size) += addend; returns new size (may grow by 1).
+constexpr std::size_t add_into_tmp(const std::span<uint_multiprecision_t>       tmp,
+                                   const std::size_t                            size,
+                                   const std::span<const uint_multiprecision_t> addend) noexcept {
+    if (addend.empty()) {
+        return size;
+    }
+
+    const std::size_t new_size = std::max(size, addend.size());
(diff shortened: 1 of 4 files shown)


include/beman/big_int/detail/mul_impl.hpp
```diff
@@ -531,23 +564,503 @@ constexpr void multiply_toom_cook_3(const std::span<uint_multiprecision_t> resul
...
+// `scratch` provides pre-allocated workspace for temporaries.
+//
+// `cutoff_override` is a benchmark-only escape hatch: when non-zero it replaces
+// `toom_cook_4_cutoff` for this call only. Production callers should omit it so
+// the default cutoff applies. Recursive sub-product calls always use the default.
+// ---------------------------------------------------------------------------
+template <class Allocator>
+constexpr void multiply_toom_cook_4(const std::span<uint_multiprecision_t> result,
+                                    std::span<const uint_multiprecision_t> a,
+                                    std::span<const uint_multiprecision_t> b,
```
eisenwave: Can we do `a_untrimmed` and `b_untrimmed` or something so we can keep these `const`?

You basically have 500 LOC below that could all potentially mutate `a` and `b`, and that's fishy. The only mutation actually taking place should be the initial trim I think.

eisenwave (changes requested):
Also might wanna sprinkle in some `[[nodiscard]]`

include/beman/big_int/detail/mul_impl.hpp
mborland: Same change has been applied to the other mul algorithms as well

eisenwave: You mean you've implemented the change here and for other multiplication algorithms? I'm not seeing anything yet on this branch.

ckormanyos:
Another fine performance enhancement. Thank you Matt. I'll stress test it, but probably after the merge...

eisenwave (approved):

PR #225, opened 2026-05-13
Trivial rng-gen seeds in BMP test harness

ckormanyos:
This PR trivially refreshes the rnd seeds when creating millions upon millions of test digits in the BMP test harness.

Hi Matt (@mborland) trivial change, but I'd like your opinion

### DIFF
--- a/tests/beman/big_int/boost_mp_testing.hpp
+++ b/tests/beman/big_int/boost_mp_testing.hpp
@@ -178,8 +178,12 @@ template <class UnaryOp>
 // that the sequence is deterministic across runs.
 // The RNG state advances on every call and is shared across all callers of this function.
 [[nodiscard]] inline std::string random_big_int(const std::size_t bits, const bool negative = false) {
+    static typename std::mt19937_64::result_type rng_seed{42};
+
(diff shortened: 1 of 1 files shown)


mborland:
Since you're starting from a fixed point and incrementing the seed you might as well keep going with the current one since we're in no danger of wrapping around a 64-bit generator. I think we either stick with 42 or use a random seed, but then we have to update all the harnesses to make sure we dump the exact values that failed rather than chasing down what the seed was.

ckormanyos:
> Since you're starting from a fixed point and incrementing the seed you might as well keep going with the current one since we're in no danger of wrapping around a 64-bit generator. I think we either stick with 42 or use a random seed, but then we have to update all the harnesses to make sure we dump the exact values that failed rather than chasing down what the seed was.

Thanks Matt.. This is now clear.

I recall from other projects we gained a slight preference for deterministic keys. So let's just stick with a `static` generator seeded up front with $42$.

In light of this conversation, I will close this PR as a not-so-great idea.

ckormanyos:
See comments above. Closing with no merge intended

PR #227, opened 2026-05-13
Instrument the Scratch Allocator and add stress testing of high-water mark

### LINKED ISSUE #224: Can we decrease memory usage of mul algorithms?
We are using 5s for Karatsuba and 6s for Toom-Cook. The former came from Boost.Multiprecision, the latter came from some back of the envelope numbers. Perhaps we can do better.

mborland:
Closes: #224


mborland:
Coverage fails due to the fact that we're not running the instrumented scratch allocator tests in CI

PR #229, opened 2026-05-14
Implement Toom-Cook 6.5

### LINKED ISSUE #228: Implement Toom 6'n'half
See: http://www.acsel-lab.com/arithmetic/arith20/papers/ARITH20_Bodrato.pdf

mborland:
Implements Toom-Cook 6.5, and adds more scientific tuning methods for our crossover points. Now these can be found in the perf/ folder. I've added a couple TODOs for myself to address repeated calls to the same function. This PR adds `shift_left_n` and `shift_right_n`, because they were super low hanging fruit and used quite a few in this function.

Closes: #228 

CC: @ckormanyos


include/beman/big_int/detail/mul_impl.hpp
```diff
@@ -1078,6 +1127,1101 @@ constexpr void multiply_toom_cook_4(const std::span<uint_multiprecision_t>
...
+    // Lambda capturing the elimination for one side (works on a, b, c, d, e where
+    // a=row-of-ones buffer, b/c=row-2/row-4 buffers, d=row-h buffer, e=row-q buffer).
+    // After return, a holds the middle coeff, b holds |d_inner| with returned sign_d_inner,
+    // c holds |d_outer| with returned sign_d_outer, d holds s_inner, e holds s_outer.
+    const auto solve_subsystem =
+        [&](std::span<uint_multiprecision_t> a_buf,
+            std::span<uint_multiprecision_t> b_buf,
+            std::span<uint_multiprecision_t> c_buf,
+            std::span<uint_multiprecision_t> d_buf,
+            std::span<uint_multiprecision_t> e_buf) constexpr noexcept -> std::pair<bool, bool> {
```
eisenwave: These parameters can be `const` I think.

eisenwave: Can we avoid `std::pair` in favor of a plain `struct` with two `bool`s for the two signs?

include/beman/big_int/detail/mul_impl.hpp
```diff
@@ -1078,6 +1127,1101 @@ constexpr void multiply_toom_cook_4(const std::span<uint_multiprecision_t>
...
+
+    // Pair (v4, vm4):  E_4 = (v4+vm4)/2;  D_4 = (v4-vm4)/8.
+    if (sign_vm4) {
+        subtract_unsigned_spans(v4, v4_view, vm4_view);
+    } else {
+        [[maybe_unused]] const bool carry_out = add_unsigned_spans(v4, v4_view, vm4_view);
+        BEMAN_BIG_INT_DEBUG_ASSERT(!carry_out);
+    }
+    {
+        [[maybe_unused]] const auto rem = shift_right_one(v4);
```
eisenwave: Is the `[[maybe_unused]]` even necessary? I thought that on release builds, it would expand into a `static_assert(requires {...})`, so the declaration still ends up being used.

include/beman/big_int/detail/mul_impl.hpp
```diff
@@ -1078,6 +1127,1101 @@ constexpr void multiply_toom_cook_4(const std::span<uint_multiprecision_t>
...
+    // -- Phase 1: symmetrize each pair. v_x <- E_x; vm_x <- D_x (with the
+    // /(2x) factor absorbed for integer-x points). Sign flags vanish after this. --
+
+    // Pair (v1, vm1):  E_1 = (v1+vm1)/2 = c0+c2+c4+c6+c8+c10;  D_1 = (v1-vm1)/2.
+    if (sign_vm1) {
+        subtract_unsigned_spans(v1, v1_view, vm1_view);
+    } else {
+        [[maybe_unused]] const bool carry_out = add_unsigned_spans(v1, v1_view, vm1_view);
+        BEMAN_BIG_INT_DEBUG_ASSERT(!carry_out);
+    }
```
eisenwave, line 1767: It seems like we are repeating this operation many times. Can we avoid the repetitiveness here?

include/beman/big_int/detail/mul_impl.hpp
```diff
@@ -1078,6 +1127,1101 @@ constexpr void multiply_toom_cook_4(const std::span<uint_multiprecision_t>
...
+    if (tmpa_size != 0 && tmpb_size != 0) {
+        multiply_toom_cook_6_5(vm1,
+                               std::span<const uint_multiprecision_t>{tmpa.data(), tmpa_size},
+                               std::span<const uint_multiprecision_t>{tmpb.data(), tmpb_size},
+                               scratch);
+    }
+
+    // ---- Evaluate at x = 2: tmpa = 32a5+16a4+8a3+4a2+2a1+a0 (Horner from a5);
+    //                         tmpb = 64b6+32b5+...+b0 (Horner from b6). ----
+    // TODO(mborland): Same thing, super repetitive calls can be handled better?
```
eisenwave, line 1349: A lot of the repetition may be avoided if there was an actual array for `a`, rather than separate variables. Then at least some of these blocks could be handled via small loop, though maybe not this one.

include/beman/big_int/detail/mul_impl.hpp
```diff
@@ -1078,6 +1127,1101 @@ constexpr void multiply_toom_cook_4(const std::span<uint_multiprecision_t>
...
+    std::ranges::copy(a5, tmpa.begin());
+    tmpa_size = a5.size();
+    tmpa_size = shift_left_one(tmpa, tmpa_size);
+    tmpa_size = add_into_tmp(tmpa, tmpa_size, a4);
```
eisenwave, line 1361: Perhaps it would already help to just fuse left-shifting and addition into a single operation using a helper lambda.

include/beman/big_int/detail/mul_impl.hpp
```diff
@@ -1112,10 +2256,19 @@ constexpr std::size_t multiply_dispatch(const std::span<uint_multiprecision_t>
         return multiply_single_limb(result, a, b[0]);
     }
 
-    // Use Toom-Cook 3 above its cutoff, Karatsuba above its cutoff, and schoolbook below.
+    // Choose algorithm to use based off the tuned cutoffs
     // Avoid these at compile time because the recursion depth could blow up consteval limits;
     // long multiplication works just fine in that case.
     if BEMAN_BIG_INT_IS_NOT_CONSTEVAL {
+        if (a.size() >= toom_cook_6_5_cutoff && b.size() >= toom_cook_6_5_cutoff) {
```
eisenwave: I haven't really paid attention to the dispatch mechanism before, and maybe I should have.

My main concern here is that you're making it go through every single unhappy case for two limbs first, but two limbs is probably more common than 2000 limbs.

It's fine if this PR piles more onto that problem, but in the long run, we want to flip the whole thing so that it checks if it should just do long multiplication first, without a critical path for that case where it goes through 8 extra size checks and 4 if statements.

eisenwave (changes requested):
Outstanding work. The performance measurements are very helpful.

I can't comment on the numeric details to be honest, this is beyond me, only one code structure.

ckormanyos:
Hi Matt I added the updated performance numbers directly to your branch

include/beman/big_int/detail/mul_impl.hpp
mborland: Seems fine without. Force of habit.

include/beman/big_int/detail/mul_impl.hpp
mborland: Marked as TODO for https://github.com/eisenwave/std-big-int/issues/231.

include/beman/big_int/detail/mul_impl.hpp
mborland: Added information to TODO and will explore as part of https://github.com/eisenwave/std-big-int/issues/231.

include/beman/big_int/detail/mul_impl.hpp
mborland: Added information to TODO and will explore as part of https://github.com/eisenwave/std-big-int/issues/231.

PR #232, opened 2026-05-15
Refactor the non-constexpr multiplication implementations into source files

### LINKED ISSUE #230: Move most of the multiplication algorithms out of the headers
We don't use anything besides school book during constant evaluation so we can shift the rest into their own source files.

mborland:
The lib is now compiled instead of header only so small updates to the CML have been made as well. The biggest change here is that the scratch allocator needed to be type erased so we can avoid the multiplication algorithms needing to be templated on the Allocator.

Closes: #230


mborland:
I'll look into reverting that last change in a future optimization PR. I'm not sure if a latent segfault became clear because of layout changes or what? Will investigate in #233

PR #236, opened 2026-05-18
Repair trivial warnings gcc-arm-none-eabi

ckormanyos:
Repair trivial compiler warnings found on `gcc-arm-none-eabi`.

### DIFF
--- a/include/beman/big_int/detail/config.hpp
+++ b/include/beman/big_int/detail/config.hpp
@@ -155,8 +155,14 @@ __extension__ using uint128_t = unsigned __int128;
 #elif defined(BEMAN_BIG_INT_MSVC)
     #define BEMAN_BIG_INT_HAS_INT128 1
     #define BEMAN_BIG_INT_HAS_INT128_CLASS 1
+    #define BEMAN_BIG_INT_HAS_INT128_FUNDAMENTAL 0
 using int128_t  = std::_Signed128;
 using uint128_t = std::_Unsigned128;
+#else
+    // Prevent warnings for use of undefined macros.
+    #define BEMAN_BIG_INT_HAS_INT128 0
+    #define BEMAN_BIG_INT_HAS_INT128_CLASS 0
+    #define BEMAN_BIG_INT_HAS_INT128_FUNDAMENTAL 0
 #endif
 
 } // namespace beman::big_int::detail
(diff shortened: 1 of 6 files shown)


ckormanyos:
This PR handles some trivial warnings I encountered on a 32-bit ARM compiler. I settled on the minimum scope of changes to work around the problem.

For a while I tried to see if it makes sense to always `#define` in all cases `BEMAN_BIG_INT_HAS_INT128`, `BEMAN_BIG_INT_HAS_INT128_CLASS` and `BEMAN_BIG_INT_HAS_INT128_FUNDAMENTAL` to be clearly 1 or 0. But I could not get that working. I'm also not sure if this would even be a good idea.

There are, however, situations where the logic seems like it _could_ be improved. But I'll not try any further for that at the moment.

include/beman/big_int/detail/config.hpp
```diff
@@ -157,6 +157,9 @@ __extension__ using uint128_t = unsigned __int128;
     #define BEMAN_BIG_INT_HAS_INT128_CLASS 1
 using int128_t  = std::_Signed128;
 using uint128_t = std::_Unsigned128;
+#else
+    // Prevent warnings for use of undefined macros.
+    #define BEMAN_BIG_INT_HAS_INT128_FUNDAMENTAL 0
```
eisenwave: `BEMAN_BIG_INT_HAS_INT128_FUNDAMENTAL` may be defined as `1` when defined just because that is somewhat idiomatic and allows direct usage in `#if`, but the intended usage in our repo is that you check #ifdef`. That's also the case with all sorts of other macros.

If you need to avoid wanrings for undefined macros, just use `#ifdef` or `#if defined`.

eisenwave (changes requested):

include/beman/big_int/detail/wide_ops.hpp
```diff
@@ -168,7 +168,7 @@ constexpr T high_mul_portable(const T x, const T y) noexcept {
 template <signed_or_unsigned T>
     requires(width_v<T> <= 64)
 constexpr T high_mul(const T x, const T y) noexcept {
-#if BEMAN_BIG_INT_HAS_INT128_FUNDAMENTAL
+#if defined(BEMAN_BIG_INT_HAS_INT128_FUNDAMENTAL) && (BEMAN_BIG_INT_HAS_INT128_FUNDAMENTAL != 0)
```
eisenwave: Just `#ifdef` is enough.

eisenwave (changes requested):

include/beman/big_int/detail/wide_ops.hpp
ckormanyos: Yup

ckormanyos:
Here is an unrelated, yet nice, result. I compiled the source-code multiplication routines both to check for warnings and get an indication of today's compiled code size. The code size is very efficient, as shown in the small report below.

It is on a 32-bit ARM(R) Cortex(R)-M4F controller.

```
+++ linking application to generate: bin/ref_app.elf
Memory region         Used Size  Region Size  %age Used
             VEC:         512 B        768 B     66.67%
             ROM:       45556 B      64768 B     70.34%
             RAM:        5780 B        28 KB     20.16%
```

include/beman/big_int/detail/config.hpp
ckormanyos: Got it

eisenwave (approved):

PR #239, opened 2026-06-03
Asym Kara cutoff and fallback

ckormanyos:
This PR investigates an asymmetric Karatsuba cutoff and fallback. Instead of $40/40$, use $48/24$.

On a few of my compilers, I obtain a not huge, but worthwhile performance improvement. I observed this phenomenon in another project and would  like to see if it makes sense here.

I need an independent confirmation of the asymmetric improvement.

I get:

| 64-bit limbs     | time mul [us] (asym)  | time mul [us] (40/40) |
|-----------------|-------------------------|------------------------|
| 48                    | 1.8                               | 1.9      |
| 128                  | 7.9                               | 8.8      |
| 256                  | 24.3                           | 27.5      |

Cc: @mborland

### DIFF
--- a/include/beman/big_int/detail/mul_impl.hpp
+++ b/include/beman/big_int/detail/mul_impl.hpp
@@ -66,7 +66,8 @@ constexpr void multiply_long(const std::span<uint_multiprecision_t>       result
 
 // Minimum number of limbs for Karatsuba to be worthwhile
 // Directly from Boost, and reconfirmed as correct
-inline constexpr std::size_t karatsuba_cutoff = 40;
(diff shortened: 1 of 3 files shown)


mborland (approved):
Seems like a good little improvement, and I can replicate the results.

ckormanyos:
> Seems like a good little improvement, and I can replicate the results.

Thanks Matt. I also edited the original post with a sanity check at $48$ limbs. So there is even a small improvement there too.

I guess the upshot of this is once you've paid the overhead of Karatsuba, remain there with recursive splitting for a longer recursion depth. It makes sense intuitively.

PR #240, opened 2026-06-04
Add special cases for squaring

mborland:
Adds a special case of each algorithm for the squaring case which we detect via a check to pointer and size. Provides a pretty uniform ~1.3x speedup for each algorithm for this case. The pointer and size check is sufficiently cheap, and should fail early on pointer compare, that it does not harm the base case of two separate values. 

CC: @ckormanyos


mborland:
Re-tuned multiplication crossover points also provide about a 10% speedup across all limb ranges.

ckormanyos:
Nice! I will redo my benchmarks.

PR #242, opened 2026-06-05
Re-enable stack based karatsuba with static assert to avoid SO

### LINKED ISSUE #233: Re-enable Karatsuba on the Stack
In #232 we started having segfaults on MSVC only with the stack region of Karatsuba. Once it was exclusively moved onto the heap everything was fine. We must have been violating our 5s buffer space at some point and it was just going unnoticed for whatever reason.

mborland:
@ckormanyos, can you try this one since you were the one that reported the initial stack overflow? I was only able to replicate by reducing the karatsuba fallback number to a point where recursion becomes non-terminating. 

Closes: #233

### DIFF
--- a/include/beman/big_int/detail/mul_impl.hpp
+++ b/include/beman/big_int/detail/mul_impl.hpp
@@ -128,7 +128,14 @@ inline constexpr std::size_t karatsuba_fallback = 40;
 // 40-4000 limbs in scratch_peak_bench) the actual peak/s ratio tops out at
 // ~3.997. 5*s leaves ~25% safety margin and matches the same generous-but-not-
 // wasteful ratio used by the Toom-Cook variants.
+static_assert(karatsuba_fallback >= 5,
+              "karatsuba_fallback < 5 makes the Karatsuba recursion non-terminating (stack overflow)");
+
 constexpr std::size_t karatsuba_storage_size(const std::size_t s) noexcept { return 5 * s; }
+
+// Maximum number of scratch limbs we are willing to place on the stack
+// Value from boost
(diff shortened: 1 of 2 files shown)


ckormanyos:
> can you try this one since you were the one that reported the initial stack overflow? I was only able to replicate by reducing the karatsuba fallback number to a point where recursion becomes non-terminating.

Hi Matt @mborland) yeah sure.

As it turns out, it looks like it is failing certain very specific tests in CI/CD right here.

I might need a day or two, but I can  try and reproduce locally on my MSVC. I'll report back when/if I make any progress.

mborland:
Yeah, I don't remember it being a problem with CI before. I'll assume that if I can get the CI green your end will also be green. I'll fire up my Windows VM to debug this one because clearly MSVC is doing something interesting.

ckormanyos:
> Yeah, I don't remember it being a problem with CI before. I'll assume that if I can get the CI green your end will also be green. I'll fire up my Windows VM to debug this one because clearly MSVC is doing something interesting.

Indeed. I'm taking the rest of today (probably) off. Then I can investigate further if you can't track it down.

mborland:
> > Yeah, I don't remember it being a problem with CI before. I'll assume that if I can get the CI green your end will also be green. I'll fire up my Windows VM to debug this one because clearly MSVC is doing something interesting.
> 
> Indeed. I'm taking the rest of today (probably) off. Then I can investigate further if you can't track it down.

Nasty little bug with one of MSVC's own attributes. [[msvc::no_unique_address]] seems to have issues compiling when the base class is non-empty, so it segfaults immediately. With it removed everything works fine. Hopefully no further problems.

PR #259, opened 2026-06-15
Add framework for GitHub Pages hosted documentation pages.

### LINKED ISSUE #258: Add GitHub Pages Site
We have the P4444 site situated which is good for the ultimate goal of the repo as a reference implementation. In addition to this we should add more standard documentation pages in an effort to make things approachable for users rather than having them read the specifications documents. Using GitHub Pages is cost-free and should be easy to maintain by setting up an auto build on push to main workflow.

mborland:
Edit: Since this will be moved to beman at some point it's easier to publish to the gh-pages branch rather than have to move a cache later

Closes: #258


ckormanyos:
See also small docs additions to the Markdowns in #260

PR #260, opened 2026-06-15
Improve docs and perf support

ckormanyos:
This PR adds a few perf files and preliminarily finishes the perf docs. Hi Matt (@mborland) could you review the final few paragraphs of docs which have very tersely included base-conversion?


ckormanyos:
See also #259 which will incorporate all this work.

mborland (approved):
Looks good to me. Once the docs page is further developed we can consolidate things there instead of an ever expanding README. For now it's certainly worth adding (and I'll just copy paste it later).

PR #289, opened 2026-06-26
Initial commit metal CI

ckormanyos:
This PR adds a bare-metal 32-bit ARM run to CI. This one will take a few recursions for paths, style, etc. It is copied from my personal real-time-cpp repo. So it is known to work as long as the paths are all OK.


ckormanyos:
This is marked as a draft PR until I work out the details.

mborland:
I'm done for the weekend so I have rebased you on top of the merges that I have been making.

mborland:
Linter should be green now as well

ckormanyos:
I struggled a bit with this one, but it is fundamentally working.
In this PR, we get a bare-metal QEMU run working on an ARM(R) Cortex(R)-M4F core.
The benchmark performs and checks a multiplication result in the Karatsuba range using custom allocator.

Why does this matter?

- The bare metal run on a 32-bit microcontroller with no OS, no I/O streams, etc. helps to verify the portability of the library in many ways.
- A ROM (program FLASH) limit has been set in the linker. So if we inadvertently add a lot of source code, this build will fail. This is a good monitor feature.

Notes:

- The build is done on the command line. So when we add files, we will need to add them manually in ths metal-YAML command.

ckormanyos:
This one is good to go

ckormanyos:
Hi Matt (@mborland) before I wrap this one up, a quick sync on your toy implementation of `ring_buffer`.

I used your `ring_buffer` from one of your recent test files. A small problem is that a default-constructed `big_int` can in certain situations be created with an allocator having a `nullptr` arena pointer.

I did a small modification to your `ring_buffer` in my metal app file. It uses a static buffer. And I adapted some constructors, etc.

I know you were not making a robust `ring_buffer`, but crashing in `multiply_into` (which is what I observed with the result variable `r`) can be a bit extreme. Your test code just so happens to avoid these types of situations.

If we do aggree on a single implementation, we could make a utility out of it.

mborland:
> Hi Matt (@mborland) before I wrap this one up, a quick sync on your toy implementation of `ring_buffer`.
> 
> I used your `ring_buffer` from one of your recent test files. A small problem is that a default-constructed `big_int` can in certain situations be created with an allocator having a `nullptr` arena pointer.
> 
> I did a small modification to your `ring_buffer` in my metal app file. It uses a static buffer. And I adapted some constructors, etc.
> 
> I know you were not making a robust `ring_buffer`, but crashing in `multiply_into` (which is what I observed with the result variable `r`) can be a bit extreme. Your test code just so happens to avoid these types of situations.
> 
> If we do aggree on a single implementation, we could make a utility out of it.

I am good with using your improved implementation. I probably should have made my ring buffer a little better for valid testing.

ckormanyos:
>> If we do aggree on a single implementation, we could make a utility out of it.

> I am good with using your improved implementation.

OK cool. I'll ensure that your tests run with my version also. Then I'll consolidate it down to a single header we can both use.

ckormanyos:
>> I am good with using your improved implementation.

> OK cool. I'll ensure that your tests run with my version also. Then I'll consolidate it down to a single header we can both use.

Done. Working and green. Merging. Thx Matt

PR #301, opened 2026-07-28
Add `LimbType` template parameter

mborland:
@eisenwave here's basically what we discussed on WhatsApp yesterday. `LimbType` is defaulted to and `static_assert` on `uint_multiprecision_t`


eisenwave (approved):
In the future, we'll probably want some kind of type alias inside `basic_big_int` so that `uint_multiprecision_t` is not used anymore in the public interface, but I'm not sure how to name that yet. `value_type` feels weird in this case, and we have so far avoided the term "limb" in the standard parts, but maybe `limb_type` should be public.

PR #304, opened 2026-07-30
Reverse the order of Limb and Allocator parameters to match P4444

mborland:


doc/modules/ROOT/pages/charconv.adoc
```diff
@@ -18,13 +18,13 @@ buffer of its own, never throws, and is usable in a constant expression.
 ----
 namespace beman::big_int {
 
-template <std::size_t b, class A, class LimbType>
+template <std::size_t b, class LimbType, class A>
```
eisenwave: While we're at it, I think we should also be consistent about how these out-of-line template parameter lists are written.

For `b` and `A`, we just use a single letter, and this is what P4444 does as well. So this should be `class L`.

`LimbType` is probably the least consistent with the rest of the standard wording. We do things like `class Hash` and `class charT`, but I don't think we spell out the `Type` as part of the name. I've been `class A` for allocators out-of-line before.

I didn't bring it up on the previous PR because I didn't have an opinion on it yet.

eisenwave (changes requested):

doc/modules/ROOT/pages/charconv.adoc
mborland: I'll drop `Type` from `LimbType` since it's written as `Limb` in the paper now as well

PR #305, opened 2026-07-31
Iterative steps school mul optimize

ckormanyos:
The purpose of this PR is to iteratively attempt low-level optimization of schoolbook multiplication for selected architectures.

The initial commit is but a first step. Code comments and annotations suggest next iterative steps.

Hi Matt (@mborland) in this first iterative step, there are several comments regarding `constexpr`/ `consteval` path. But these might be unecessary, since this routine is used from Karatsuba as a fallback? Could you please comment on this at a time of your convenience?

Cc: @eisenwave


include/beman/big_int/detail/multiply_long_runtime.hpp
```diff
@@ -0,0 +1,70 @@
...
+//                   uncommenting one of the following two macros.
+//                   When the upper macro is active, assembly is used.
+//                   When the lower macro is active, C++ code is used.
+//                   Only one macro should be activated at any time.
+//                   Note that the upper macro spans multiple lines.
+
+#define MULTIPLY_LONG_RUNTIME(PARAM_RESULT, PARAM_A, PARAM_B) \
+    ::multiply_long_runtime(                                  \
+        (PARAM_RESULT).data(), (PARAM_A).data(), (PARAM_A).size(), (PARAM_B).data(), (PARAM_B).size())
+// #define MULTIPLY_LONG_RUNTIME(PARAM_RESULT, PARAM_A, PARAM_B) multiply_long((PARAM_RESULT), (PARAM_A), (PARAM_B))
```
eisenwave: If this is unconditionally calling a regular function, it seems like this doesn't need to be a macro at all.

include/beman/big_int/detail/multiply_long_runtime.hpp
```diff
@@ -0,0 +1,70 @@
...
+                                      const beman::big_int::uint_multiprecision_t* p_b,
+                                      const std::size_t                            len_b) noexcept;
+
+#else
+
+extern "C" inline void multiply_long_runtime(beman::big_int::uint_multiprecision_t*       p_result,
+                                             const beman::big_int::uint_multiprecision_t* p_a,
+                                             const std::size_t                            len_a,
+                                             const beman::big_int::uint_multiprecision_t* p_b,
+                                             const std::size_t                            len_b) noexcept {
```
eisenwave: Since the signature is the same in both function heads, can you make it so only the function body and the `inline` keyword are `#ifdef`d?

include/beman/big_int/detail/multiply_long_runtime.hpp
```diff
@@ -0,0 +1,70 @@
...
+#define MULTIPLY_LONG_RUNTIME(PARAM_RESULT, PARAM_A, PARAM_B) \
+    ::multiply_long_runtime(                                  \
+        (PARAM_RESULT).data(), (PARAM_A).data(), (PARAM_A).size(), (PARAM_B).data(), (PARAM_B).size())
+// #define MULTIPLY_LONG_RUNTIME(PARAM_RESULT, PARAM_A, PARAM_B) multiply_long((PARAM_RESULT), (PARAM_A), (PARAM_B))
+
+BEMAN_BIG_INT_DIAGNOSTIC_POP()
+
+#if defined(BEMAN_BIG_INT_ARCH_X86_64_GCC_OR_CLANG)
+
+extern "C" void multiply_long_runtime(beman::big_int::uint_multiprecision_t*       p_result,
```
eisenwave: We shouldn't pollute the global namespace without some library prefix. It should be something like `beman_big_int_multiply_long_runtime`.

eisenwave (changes requested):

include/beman/big_int/detail/multiply_long_runtime.hpp
ckormanyos: Hi Jan (@eisenwave) the macro will be dropped toward the upcoming end of the PR. Its sole purpose right now is to activate/deactivate the assembly optimization, switching to/from the asm/C++ implementation so that we can do direct comparisons locally.

include/beman/big_int/detail/multiply_long_runtime.hpp
ckormanyos: Agreed

include/beman/big_int/detail/multiply_long_runtime.hpp
ckormanyos: Agreed

PR #318, opened 2026-08-17
Implement, test, and document GCD

### LINKED ISSUE #320: Implement GCD

mborland:
CC: @ckormanyos

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -263,9 +263,13 @@ template <class T>
     requires detail::is_basic_big_int_v<T>
 constexpr std::remove_cvref_t<T> abs(T&& x);
 
+// `gcd` itself is a set of overloads in <beman/big_int/numeric.hpp>; they all run
+// through this driver, which is the one that needs the representation.
+namespace detail {
 template <class M, class N>
-    requires detail::common_big_int_type_with<M, N>
-constexpr detail::common_big_int_type<M, N> gcd(M m, N n);
+    requires common_big_int_type_with<M, N>
+constexpr common_big_int_type<M, N> gcd_impl(M&& m, N&& n);
+} // namespace detail
 
 // [big.int.class], class template basic_big_int
 template <std::size_t min_inplace_bits, class Limb, class Allocator>
@@ -566,7 +570,7 @@ class BEMAN_BIG_INT_TRIVIAL_ABI basic_big_int {
 
     template <class M, class N>
         requires detail::common_big_int_type_with<M, N>
-    friend constexpr detail::common_big_int_type<M, N> gcd(M m, N n);
+    friend constexpr detail::common_big_int_type<M, N> detail::gcd_impl(M&& m, N&& n);
 
   private:
     template <detail::unsigned_integer T>
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -263,6 +263,10 @@ template <class T>
     requires detail::is_basic_big_int_v<T>
 constexpr std::remove_cvref_t<T> abs(T&& x);
 
+template <class M, class N>
+    requires detail::common_big_int_type_with<M, N>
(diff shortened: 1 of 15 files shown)


include/beman/big_int/detail/config.hpp
```diff
@@ -136,6 +136,12 @@ using bit_uint = unsigned _BitInt(N);
     #define BEMAN_BIG_INT_BITINT_MAXWIDTH 0
 #endif // BITINT_MAXWIDTH
 
+// Workaround for Clang-19 ICE past 128 bits, even though it reports far more than that
```
eisenwave: Can you add a link to an LLVM issue here so we know when it's fixed and what this ICE is about?

include/beman/big_int/detail/gcd_impl.hpp
```diff
@@ -0,0 +1,336 @@
...
+// larger with their difference. Division-free, so this is the fast path for
+// every pair of operands that fits a limb -- including the tail of the
+// multi-limb algorithm below.
+// Both operands must be nonzero.
+[[nodiscard]] constexpr uint_multiprecision_t gcd_limbs(uint_multiprecision_t u, uint_multiprecision_t v) noexcept {
+    BEMAN_BIG_INT_DEBUG_ASSERT(u != 0);
+    BEMAN_BIG_INT_DEBUG_ASSERT(v != 0);
+
+    const auto common = static_cast<unsigned>(std::countr_zero(u | v));
+    u >>= static_cast<unsigned>(std::countr_zero(u));
```
eisenwave: These `static_cast`s aren't necessary, are they? There is no implicit conversion that would otherwise take place in `>>=`.

include/beman/big_int/numeric.hpp
```diff
@@ -55,6 +61,176 @@ constexpr R saturating_cast(const basic_big_int<b, L, A>& x) noexcept {
...
+// argument magnitude can fail to be representable. The result is never negative:
+// gcd(m, n) == gcd(|m|, |n|), and gcd(0, 0) is 0.
+//
+// The parameters are taken by value, matching `std::gcd`: the signature makes this
+// overload the more constrained of the two, so an unqualified call with a
+// `basic_big_int` argument is not ambiguous with `std::gcd`. Pass an rvalue to hand
+// a big_int argument's storage over to the algorithm.
+template <class M, class N>
+    requires detail::common_big_int_type_with<M, N>
+[[nodiscard]] constexpr detail::common_big_int_type<M, N> gcd(M m, N n) {
```
eisenwave: This should be accepting two forwarding references and then run the same case classification into `binary_op_form` as `operator+`.

eisenwave (changes requested):

include/beman/big_int/big_int.hpp
```diff
@@ -263,6 +263,10 @@ template <class T>
     requires detail::is_basic_big_int_v<T>
 constexpr std::remove_cvref_t<T> abs(T&& x);
 
+template <class M, class N>
+    requires detail::common_big_int_type_with<M, N>
```
eisenwave: I don't think the requires is needed. SFINAE on the return type should handle it.

include/beman/big_int/detail/config.hpp
mborland: Added a related PR, but there seems to be no open issue. This is fixed as of Clang 20.

include/beman/big_int/detail/gcd_impl.hpp
mborland: These will warn on sign conversion without as `std::countr_zero` unfortunately returns an `int`.

eisenwave:
You've marked the comment as resolved but I can still see some `requires detail::common_big_int_type_with<M, N>`, so is that actually necessary?

mborland:
> You've marked the comment as resolved but I can still see some `requires detail::common_big_int_type_with<M, N>`, so is that actually necessary?

I'm pulling it off `gcd_impl` now, I missed that one when it became the impl function and added the driver functions like the binary ops.

PR #332, opened 2026-08-19
Fix the non-canonical magnitude left by decrement across a limb boundary

### LINKED ISSUE #330: `--` and `++` across a limb boundary leave a non-canonical value, so `operator>>` returns 0 and `std::hash` breaks its invariant
### Describe the deficiency

Investigation was AI-assisted; the reproduction below was run on two compilers.

`operator--` and `operator++` can leave a `basic_big_int` whose top limb is zero, which as I understand it violates the trimmed-top-limb invariant named at `big_int.hpp:623`. Comparison still reports the right value, so it is easy to miss, but several observers do not:

- the free `operator>>` **loses the value outright**. `width_mag()` reads 0 on a de-normalized magnitude, and the early exit at `big_int.hpp:2085` discards everything at or beyond that width, so `x >> 0` returns `0` for a value of 2^128 - 1. Compound `>>=` is correct, because it calls `shift_right()` and never consults `width_mag()`.
- `to_string`, `to_chars`, `std::format` and base-16 `to_string` all print `0`.
- `size()` and `representation_size()` return 0, and `representation_size()` disagreeing with `representation().size()` contradicts the contract documented at `big_int.hpp:1369`.
- `std::hash` returns different values for two objects that compare equal, so an `unordered_map` or `unordered_set` keyed on `big_int` can fail to find a key that is present.
- `shrink_to_fit()` cannot recover: the value fits one limb but the stored count says two, so it takes its early return and stays on the heap.

It reproduces at limb-aligned powers of two (2^64, 2^128, 2^192) and is clean either side of them (2^63, 2^65, 2^127, 2^129). `y -= 1` and `y - 1` are both correct; only the increment and decrement operators are affected. `operator~` is affected too, since it is `negate(); --copy;`, which is how I came across it.

A guess at the root cause, which you will be able to confirm or dismiss much faster than I can: `unchecked_decrement_magnitude()` (`include/beman/big_int/big_int.hpp:1230-1247`) propagates the borrow and then handles only the borrow-out case, correctly noting that can happen only when the magnitude was zero, but it does not appear to trim a top limb that the borrow zeroed. Decrementing 2^64 turns limbs `[0, 1]` into `[0xFFFFFFFFFFFFFFFF, 0]` and leaves `limb_count()` at 2. The sibling `unchecked_increment_magnitude()` at `:1213-1227` does handle its mirror case, calling `reserve_representation` and extending the count on carry-out, so growth looks handled and shrinking does not.

### To Reproduce

```c++
#include <beman/big_int/big_int.hpp>
#include <beman/big_int/string.hpp>
#include <cassert>
#include <cstdio>
#include <functional>

int main() {
    using beman::big_int::big_int;

    const big_int a = big_int{1} << 64;
    big_int       y = a;
    --y;                             // y should be 2^64 - 1
    const big_int ref = a - big_int{1};

    assert(y == ref);                // passes: comparison is correct
    std::printf("--   : %s\n", to_string(y).c_str());
    std::printf("- 1  : %s\n", to_string(ref).c_str());

    const std::hash<big_int> h;
    std::printf("hash equal: %s\n", h(y) == h(ref) ? "yes" : "no");
}
```

Output on both GCC 13.3.0 and Clang 21.1.8:

```
--   : 0
- 1  : 18446744073709551615
hash equal: no
```

At 2^128 the value loss through `operator>>` is visible directly: `x >> 0` gives `0` where `(power - 1) >> 0` gives `340282366920938463463374607431768211455`, while `x >>= 1` on the same object is correct.

### Expected Behavior

```
--   : 18446744073709551615
- 1  : 18446744073709551615
hash equal: yes
```

`--y` producing the same representation as `a - big_int{1}`, and `std::hash` agreeing for values that compare equal, per [hash.requirements].

### Additional Discussions

What made this awkward to pin down is that the non-canonical object behaves inconsistently rather than uniformly badly. Value predicates all look correct: `y == ref`, `y != 0`, `y > 0` and `y <=> ref` all gave the right answer. The serializers are all wrong: `to_string`, `std::format("{}")`, `to_chars` and base-16 `to_string` produced `0`, and `width_mag()` returned 0 where the correctly built equal value returns 64. Adding or subtracting zero preserved the broken state, while `* 1` and `*= 1` repaired it, which I assume is because the multiply path recomputes the limb count.

Not memory unsafety, as far as I can tell. Clean under ASan and UBSan via the `llvm-debug` preset, so this looks purely like normalization logic.

The behaviour appears symmetric on the negative side, since incrementing a negative value decrements the magnitude, so `++` on `-2^64` fails the same way.

Both suites pass at this commit (933 tests on `gcc-debug`, 950 on `llvm-debug`), so increment and decrement across a limb boundary looks more like a coverage gap than a known issue. If that is right, extending the Boost.Multiprecision differential oracle to `++`, `--` and `~` with limb-aligned operands would probably catch this class of thing, though you are better placed than I am to judge where that fits.

I have a test case and a candidate one-line fix, and would be glad to open a PR with both if that is welcome. The fix normalizes after the borrow in `unchecked_decrement_magnitude()` using the existing `unchecked_trim_magnitude()` helper. The tests extend `increment_decrement.test.cpp` and `hash.test.cpp`, cover prefix and postfix `--`, `++` on negatives, both `operator~` overloads, the in-place path through `basic_big_int<256>`, the magnitude observers above, and `shrink_to_fit()`, each at 2^64, 2^128 and 2^192. Worth noting that the invariant is also checkable at compile time: putting the decrement inside a `consteval` function keeps the transient allocation inside the constant expression, so a `static_assert` catches this without the fix. With the fix, `gcc-debug` is 941/941 and `llvm-debug` 958/958, both warning-clean, and `pre-commit run --all-files` is green.

Versions: commit `84885a40`, GCC 13.3.0 with the `gcc-debug` preset and Clang 21.1.8 with the `llvm-debug` preset. Both reproduce identically.

xroche:
Fix and tests were AI-assisted; I ran both debug presets and `pre-commit` locally.

Closes: #330.

`unchecked_decrement_magnitude` did not trim a top limb the borrow had zeroed.

The tests assert the representation rather than the value, because asserting the value is what let this through: `PrefixDecrementAllocatedBorrowChain` already covered the failing case and passed. `is_normalized` went into `testing.hpp` because `hash.test.cpp` needs it too.

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -1242,6 +1242,8 @@ constexpr bool basic_big_int<b, L, A>::unchecked_decrement_magnitude() {
         BEMAN_BIG_INT_DEBUG_ASSERT(limb_count() != 0);
         limbs[0] = 1;
         unchecked_set_limb_count(1);
+    } else {
+        unchecked_trim_magnitude();
     }
     return borrow_in;
 }
--- a/tests/beman/big_int/hash.test.cpp
+++ b/tests/beman/big_int/hash.test.cpp
@@ -358,4 +358,18 @@ TEST(Hash, BoundarySignedInt64Min) {
     EXPECT_EQ(h(i64_min), h(i64_min_copy));
 }
 
+TEST(Hash, DecrementAcrossLimbBoundaryPreservesHash) {
+    const std::hash<big_int> hasher;
+    for (const unsigned shift : {64U, 128U, 192U}) {
+        big_int x = big_int{1} << shift;
+        --x;
+
+        const big_int expected = (big_int{1} << shift) - big_int{1};
+        ASSERT_EQ(x, expected) << "shift " << shift;
+
+        EXPECT_TRUE(is_normalized(x)) << "shift " << shift;
+        EXPECT_EQ(hasher(x), hasher(expected)) << "shift " << shift;
+    }
+}
+
 } // namespace
--- a/tests/beman/big_int/increment_decrement.test.cpp
+++ b/tests/beman/big_int/increment_decrement.test.cpp
@@ -2,7 +2,9 @@
 // SPDX-License-Identifier: BSL-1.0
 
 #include <compare>
+#include <cstddef>
 #include <limits>
+#include <string>
 
 #include <gtest/gtest.h>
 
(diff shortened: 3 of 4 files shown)


mborland (approved):
This looks correct to me. Thank you!

PR #333, opened 2026-08-19
Make hash independent of limb width 

### LINKED ISSUE #331: `std::hash` depends on the limb width, so equal values hash differently on 32-bit targets
### Describe the deficiency

Investigation was AI-assisted; every number below is measured.

P4444's `[big.int.hash]` requires the hash to depend only on the integer value, and the paper's author confirmed in eisenwave/cpp-proposals#242 that this is meant to hold irrespective of limb type. `detail::siphash` currently hashes the limb sequence rather than the value, so it does not meet that requirement.

A 32-bit target selects 32-bit limbs by default, with no macro involved (`INTPTR_MAX == INT32_MAX` leads to `BEMAN_BIG_INT_LIMB_WIDTH 32`), and the `metal` workflow builds and runs on stm32f429 on every push. So this is a cross-platform hash-stability gap rather than a hypothetical one.

The digest is a function of the limb decomposition in two ways that always occur together, both governed by the parity of the 32-bit limb count. `byte_len` is `limbs.size() * sizeof(uint_multiprecision_t)`, and under 32-bit limbs an odd trailing limb folds into the final block instead of being compressed as a word. For the value 5:

| | 64-bit limbs | 32-bit limbs |
|---|---|---|
| `limbs` | `[5]` | `[5]` |
| `byte_len` | `1 * 8 = 8` | `1 * 4 = 4` |
| words compressed | 1, `0x0000000000000005` | 0, since `pairs = 1/2 = 0` |
| final block | `0x0800000000000000` | `0x0400000000000005` |

Because both effects follow from the same condition, the two configurations agree exactly when the 32-bit limb count is even, which is when `byte_len` matches. In value terms they agree when `bit_width(abs(x))` lies in `(64k+32, 64k+64]` and disagree otherwise, including for zero, whose representation is one limb.

### To Reproduce

```
g++ -std=c++23 -DBEMAN_BIG_INT_FORCED_LIMB_WIDTH=64 -I include hash.cpp -o h64
g++ -std=c++23 -DBEMAN_BIG_INT_FORCED_LIMB_WIDTH=32 -I include hash.cpp -o h32
```

Constructing a `big_int` and calling `std::hash` needs no linking against the library, so neither binary picks up the other configuration's kernels.

`std::hash` values, GCC 13.3.0, x86-64, 10 of 24 values agreeing:

| value | 64-bit limbs | 32-bit limbs | |
|---|---|---|---|
| 0 | `39d3851ca07681a7` | `eff099d7f6039771` | differ |
| 1 | `2b91b2b085e6d1f6` | `00052ba70546f7c4` | differ |
| 5 | `094db73b83961848` | `dd1b37ddbda8102f` | differ |
| 12345 | `4a82911729d15a3e` | `3dd0072230f98af1` | differ |
| 256 | `b7509dad048f9b95` | `c01b5a351a432776` | differ |
| 2^32-1 | `68b88f4caf658465` | `39dc9ede851c3408` | differ |
| 2^32 | `6e5ece299655b9cb` | `6e5ece299655b9cb` | agree |
| 2^40 | `6853aa316dc926e0` | `6853aa316dc926e0` | agree |
| 2^64-1 | `2a68ff30a3d9da34` | `2a68ff30a3d9da34` | agree |
| 2^64 | `2fa2d16f0e15bf11` | `573459094a77c1cf` | differ |
| 2^128-1 | `a06ef3ea6d6b622e` | `a06ef3ea6d6b622e` | agree |
| 2^128 | `6524447f385b5ae4` | `c2d4489c88f9f947` | differ |

One caveat I should state rather than leave you to find: the forced-macro build is a faithful proxy for this code path but not for the library as a whole. `siphash()` branches only on `sizeof(uint_multiprecision_t)`, which is 4 both under the macro and on a real 32-bit target. The library target builds clean with the macro, but the test suite does not: three sites trip `-Werror` under `-Wconversion`, `-Wpadded` and `-Woverflow`, and all three come from the combination of 32-bit limbs with a 64-bit `size_t` that no real target has.

### Expected Behavior

Equal values hash equally regardless of limb width. `hash.test.cpp` already pins this for one of the three template parameters, in `CrossInplaceBitsForSmallValue`, `CrossInplaceBitsForMultiLimb` and `CrossInplaceBitsForZero`. There is no cross-limb-type analogue, and the limb type is the only one of the three parameters that changes the digest: `basic_big_int<64>`, `basic_big_int<256>` and `pmr::basic_big_int<64>` all agree.

### Additional Discussions

A patch is below if the direction is welcome. It hashes the little-endian byte string of the magnitude, trimmed of its most significant zero bytes, instead of the limb sequence. That makes the digest a function of the value by construction.

Trimming has to mean dropping only the most significant zero bytes, so the length is `ceil(bit_width(abs(x))/8)` and zero maps to the empty string; 256 keeps both of its bytes. That is the rule `significant_byte_len` already applies in the test support so that magnitudes compare equal to Boost.Multiprecision regardless of limb width.

Measured with the patch:

- identical digests under both limb widths for all 24 values above, up from 10.
- `gcc-debug` 933/933 and `llvm-debug` 950/950, both warning-clean, and `pre-commit run --all-files` green.
- `hash(0)` becomes `726fdb47dd0e0e31`, which is the SipHash-2-4 reference vector for the empty input. Nothing in the suite pins a digest, so no test needed updating.
- hash cost is unchanged within noise for values of 64 bits and wider (11.4 against 10.4 ns at 64 bits, 13.5 against 12.9 at 256, 138.7 against 138.7 at 4096). Values narrower than one limb cost about 3 ns more, because they take the byte-wise tail path.

Three things worth knowing before you decide.

It changes every hash value. That is cheap now and expensive after standardization, so the timing argues for doing it sooner rather than later, but it is your call.

A narrower fix does not work. Deriving `byte_len` from the significant bytes while keeping the limb loops still diverges: for the value 5 the 64-bit path compresses a word and leaves the final block empty, while the 32-bit path compresses nothing and folds 5 into the final block.

It interacts with #330. A hash over significant bytes ignores a trailing zero limb, so it would mask the `std::hash` symptom in that issue without fixing the underlying normalization bug, where `operator>>` still returns 0. If both changes land, the durable assertions in #330 are the ones on `representation()`, `width_mag()` and `operator>>`, not the hash one.

```diff
diff --git a/include/beman/big_int/detail/siphash.hpp b/include/beman/big_int/detail/siphash.hpp
index 2be9fa4..3072595 100644
--- a/include/beman/big_int/detail/siphash.hpp
+++ b/include/beman/big_int/detail/siphash.hpp
@@ -71,30 +71,46 @@ constexpr std::size_t siphash(const std::span<const uint_multiprecision_t> limbs
         s.v3 ^= std::numeric_limits<std::uint64_t>::max();
     }
 
-    const auto byte_len = limbs.size() * sizeof(uint_multiprecision_t);
+    // SipHash-2-4 over the little-endian byte string of the magnitude, trimmed
+    // of its most significant zero bytes. Feeding bytes rather than limbs keeps
+    // the digest a function of the value, so equal values hash equally across
+    // limb types as [big.int.hash] requires. Same trimming rule as
+    // significant_byte_len in the test support.
+    constexpr std::size_t limb_bytes = sizeof(uint_multiprecision_t);
+
+    const auto byte_of = [limbs](const std::size_t i) -> std::uint64_t {
+        const auto shift = static_cast<unsigned>((i % limb_bytes) * 8U);
+        return static_cast<std::uint64_t>(static_cast<std::uint8_t>(limbs[i / limb_bytes] >> shift));
+    };
 
-    if constexpr (sizeof(uint_multiprecision_t) == sizeof(std::uint64_t)) {
-        for (const auto m : limbs) {
-            s = compress(s, static_cast<std::uint64_t>(m));
-        }
-    } else {
-        // Pack pairs of uint32_t into uint64_t words
-        const auto pairs = limbs.size() / 2;
-        for (std::size_t i = 0; i < pairs; ++i) {
-            auto m = static_cast<std::uint64_t>(limbs[i * 2]) | (static_cast<std::uint64_t>(limbs[i * 2 + 1]) << 32);
-            s      = compress(s, m);
-        }
-        // Odd trailing uint_multiprecision_t folds into the final block
+    std::size_t byte_len = limbs.size() * limb_bytes;
+    while (byte_len != 0U && byte_of(byte_len - 1U) == 0U) {
+        --byte_len;
     }
 
-    // Final block: length in high byte + possible trailing uint32_t
-    auto b = static_cast<std::uint64_t>(byte_len) << 56;
-    if constexpr (sizeof(uint_multiprecision_t) == sizeof(std::uint32_t)) {
-        if (limbs.size() & 1) {
-            b |= static_cast<std::uint64_t>(limbs.back());
+    // A whole 8-byte block is one 64-bit limb, or a pair of 32-bit limbs, so the
+    // full blocks need no byte-wise assembly. Only the trailing partial block does.
+    const std::size_t blocks = byte_len / 8U;
+    for (std::size_t k = 0; k != blocks; ++k) {
+        if constexpr (limb_bytes == sizeof(std::uint64_t)) {
+            s = compress(s, static_cast<std::uint64_t>(limbs[k]));
+        } else {
+            const auto lo = static_cast<std::uint64_t>(limbs[k * 2U]);
+            const auto hi = static_cast<std::uint64_t>(limbs[k * 2U + 1U]);
+            s             = compress(s, lo | (hi << 32U));
         }
     }
 
+    // Final block: length in the high byte, then the trailing partial word.
+    // The cast is needed where size_t is narrower than uint64_t.
+    BEMAN_BIG_INT_DIAGNOSTIC_PUSH()
+    BEMAN_BIG_INT_DIAGNOSTIC_IGNORED_GCC("-Wuseless-cast")
+    auto b = static_cast<std::uint64_t>(byte_len & 0xFFU) << 56U;
+    BEMAN_BIG_INT_DIAGNOSTIC_POP()
+    for (std::size_t j = 0; j != byte_len % 8U; ++j) {
+        b |= byte_of(blocks * 8U + j) << (j * 8U);
+    }
+
     s = compress(s, b);
 
     s.v2 ^= 0xFFULL;
```

mborland:
Closes: #331

### DIFF
--- a/include/beman/big_int/detail/siphash.hpp
+++ b/include/beman/big_int/detail/siphash.hpp
@@ -50,6 +50,18 @@ constexpr state_holder compress(state_holder s, const std::uint64_t m) {
     return s;
 }
 
+// Length of the little-endian byte string of the magnitude `limbs`, trimmed of its
+// most significant zero bytes. That is ceil(bit_width(magnitude) / 8), which does not
+// depend on how the magnitude is split into limbs. A zero magnitude has length zero.
+constexpr std::size_t significant_byte_len(const std::span<const uint_multiprecision_t> limbs) noexcept {
+    for (std::size_t i = limbs.size(); i != 0; --i) {
+        if (const auto top = limbs[i - 1]; top != 0) {
+            return (i - 1) * sizeof(uint_multiprecision_t) + static_cast<std::size_t>((std::bit_width(top) + 7) / 8);
+        }
+    }
+    return 0;
+}
+
 } // namespace impl
 
 constexpr std::size_t siphash(const std::span<const uint_multiprecision_t> limbs, const bool sign) {
@@ -71,29 +83,34 @@ constexpr std::size_t siphash(const std::span<const uint_multiprecision_t> limbs
         s.v3 ^= std::numeric_limits<std::uint64_t>::max();
     }
 
-    const auto byte_len = limbs.size() * sizeof(uint_multiprecision_t);
+    // SipHash-2-4 over the little-endian byte string of the magnitude, trimmed of its
+    // most significant zero bytes.
+    constexpr std::size_t limb_bytes = sizeof(uint_multiprecision_t);
 
-    if constexpr (sizeof(uint_multiprecision_t) == sizeof(std::uint64_t)) {
-        for (const auto m : limbs) {
-            s = compress(s, static_cast<std::uint64_t>(m));
+    const auto byte_len = impl::significant_byte_len(limbs);
+
+    const auto limb_at = [limbs](const std::size_t i) -> std::uint64_t {
+        return i < limbs.size() ? static_cast<std::uint64_t>(limbs[i]) : std::uint64_t{0};
+    };
+
+    // Block k of that byte string
+    const auto block_at = [&limb_at](const std::size_t k) -> std::uint64_t {
+        if constexpr (limb_bytes == sizeof(std::uint64_t)) {
+            return limb_at(k);
+        } else {
+            return limb_at(k * 2) | (limb_at(k * 2 + 1) << 32U);
         }
-    } else {
-        // Pack pairs of uint32_t into uint64_t words
-        const auto pairs = limbs.size() / 2;
-        for (std::size_t i = 0; i < pairs; ++i) {
-            auto m = static_cast<std::uint64_t>(limbs[i * 2]) | (static_cast<std::uint64_t>(limbs[i * 2 + 1]) << 32);
-            s      = compress(s, m);
-        }
-        // Odd trailing uint_multiprecision_t folds into the final block
+    };
+
+    const auto blocks = byte_len / 8;
+    for (std::size_t k = 0; k != blocks; ++k) {
+        s = compress(s, block_at(k));
     }
 
-    // Final block: length in high byte + possible trailing uint32_t
-    auto b = static_cast<std::uint64_t>(byte_len) << 56;
-    if constexpr (sizeof(uint_multiprecision_t) == sizeof(std::uint32_t)) {
-        if (limbs.size() & 1) {
-            b |= static_cast<std::uint64_t>(limbs.back());
-        }
-    }
+    // Final block: length in the high byte, then the trailing partial block, which is
+    // the low byte_len % 8 bytes of the block past the last full one.
+    const auto tail_mask = (std::uint64_t{1} << ((byte_len % 8) * 8)) - 1;
+    const auto b         = (static_cast<std::uint64_t>(byte_len) << 56) | (block_at(blocks) & tail_mask);
 
     s = compress(s, b);
 
--- a/include/beman/big_int/detail/siphash.hpp
+++ b/include/beman/big_int/detail/siphash.hpp
@@ -21,7 +21,7 @@ struct state_holder {
 
 namespace impl {
 
-constexpr state_holder sipround(state_holder s) {
+[[nodiscard]] constexpr state_holder sipround(state_holder s) {
     s.v0 += s.v1;
     s.v1 = std::rotl(s.v1, 13);
     s.v1 ^= s.v0;
@@ -42,7 +42,7 @@ constexpr state_holder sipround(state_holder s) {
     return s;
 }
 
-constexpr state_holder compress(state_holder s, const std::uint64_t m) {
+[[nodiscard]] constexpr state_holder compress(state_holder s, const std::uint64_t m) {
     s.v3 ^= m;
     s = sipround(s);
     s = sipround(s);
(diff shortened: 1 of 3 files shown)


include/beman/big_int/detail/siphash.hpp
```diff
@@ -50,6 +50,18 @@ constexpr state_holder compress(state_holder s, const std::uint64_t m) {
     return s;
 }
 
+// Length of the little-endian byte string of the magnitude `limbs`, trimmed of its
+// most significant zero bytes. That is ceil(bit_width(magnitude) / 8), which does not
+// depend on how the magnitude is split into limbs. A zero magnitude has length zero.
+constexpr std::size_t significant_byte_len(const std::span<const uint_multiprecision_t> limbs) noexcept {
```
eisenwave: While we're at it, this and `sighash` could use `[[nodiscard]]`.

include/beman/big_int/detail/siphash.hpp
```diff
@@ -50,6 +50,18 @@ constexpr state_holder compress(state_holder s, const std::uint64_t m) {
...
+    for (std::size_t i = limbs.size(); i != 0; --i) {
+        if (const auto top = limbs[i - 1]; top != 0) {
+            return (i - 1) * sizeof(uint_multiprecision_t) + static_cast<std::size_t>((std::bit_width(top) + 7) / 8);
```
eisenwave: That division looks like a `detail::div_to_pos_inf` to me.

eisenwave (approved):
Tests and hash code look good, just some stylistic nitpicks you could deal with if you feel like it.

PR #337, opened 2026-08-19
Change hash to match `std::hash<_BitInt(N)>(static_cast<_BitInt(N)>(x))`

mborland:
I think we've questioned whether or not to do this a few times. I think this follows since if we have two values that compare equal, their hashes will also compare equal. I'll label it as a question; I just happened to be thinking about it again because of our earlier hash related issue.

### DIFF
--- a/.github/workflows/metal.yml
+++ b/.github/workflows/metal.yml
@@ -19,7 +19,6 @@ jobs:
           fetch-depth: '0'
       - name: update-tools
         run: |
-          sudo apt update
           sudo apt install libncurses5-dev
           mkdir -p emu_env && cd emu_env
           wget --no-check-certificate https://gitlab.arm.com/api/v4/projects/tooling%2Fgnu-toolchains-for-arm/packages/generic/gnu-toolchain/15.3.rel1/arm-gnu-toolchain-15.3.rel1-x86_64-arm-none-eabi.tar.xz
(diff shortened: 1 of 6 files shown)


eisenwave (changes requested):
Modern standard libraries support `std::hash<__int128>`, and libc++ supports up to `std::hash<_BitInt(256)>`. If we want to guarantee parity with integer types, `int64_t` as a cutoff point fundamentally doesn't make sense.

We should ensure that at least if those are supported, our tests cover parity for `__int128`, and I guess we need to do some digging and see what hashing these standard libraries are doing for wide integers.

mborland:
`__int128` like all other integer types is hashed as `static_cast<size_t>()`.  I can see if `_BitInt` can be made to work on Clang and just ladder by increasingly wide type.

eisenwave (changes requested):
This looks mostly good, but may need some refactoring if much wide `_BitInt` hashing starts to be supported by libc++.

The one thing I feel is missing is a test that goes through all the `_BitInt` widths from 1 to 256 and makes sure we have hash parity with those. It seems like the tests only cover multiples of 64.

ckormanyos:
Hi @mborland the sporadic CI-metal failures were at the point of `sudo apt update`.

The metal run installs a few packages. But I don`t think the update is needed. Down the road, I'd like to make an image for these kinds of runs. You need a modern `arm-none-eabi-gcc`, `libncurses5-dev` and an xpack emulator. At least that's the toolset that i have collected to run QEMU via GDB. To get more, let's say, permanent, I'd want to make an image later.

For now, I reran the metal run and it was fine. I will also remove the update and see how it runs.

eisenwave (approved):

PR #338, opened 2026-08-19
Make assignment strong against a throwing allocation

### LINKED ISSUE #335: Copy-assignment leaves the object owning freed storage if the allocation throws
Investigation was AI-assisted; every outcome below was measured on `79f71eb` with clang-21 and g++-13, both under ASan and LSan.

Same-type copy-assignment leaves the object owning storage it has already freed, if the allocation throws. `assign_value`'s slow path frees first and allocates second:

```cpp
free_storage();                       // big_int.hpp:729
m_size_and_sign = src.m_size_and_sign;   // :730
// ...
const alloc_result allocation = alloc_limbs(needed);   // :777  throws here
copy_n_to_allocation(src.limb_ptr(), src_count, allocation);
m_capacity     = static_cast<std::uint32_t>(allocation.count);   // :779
m_storage.data = allocation.ptr;                                 // :780
```

Lines 779 and 780 are the only place `m_capacity` and `m_storage.data` are corrected, and they sit below the throw. So after `bad_alloc` escapes, the control words still describe the released block.

`grow()` at `:3329` already does it in the safe order: `alloc_limbs`, `copy_n_to_allocation`, then `free_storage()`. So does `+=` (`:2437`/`:2447`), `-=` (`:2524`/`:2534`, `:2553`/`:2563`), and both branches of `shrink_to_fit` (`:1454`, `:1459`). `assign_value:729` is the only place in the header where the free precedes the allocation, and it is exactly the operation that measures unsafe while `+=` and `-=` measure strong.

Two failure modes, depending on where the destination's magnitude lives. With `x` on the heap (`x = big_int{1} << 100`, `w = big_int{1} << 4000`), `x` ends up owning a freed block: `to_string(x)` gives `heap-buffer-overflow` reading in `width_mag()` at `:1356`, because the stale limb count runs past the freed allocation; a read starting at limb 0 such as the copy constructor gives `heap-use-after-free`; and letting the scope exit without touching `x` at all gives `attempting double-free` from `~basic_big_int` at `:1179`.

With `x` still inline (`x{7}`) there is no dangling pointer, but `m_capacity` stays 0 while the limb count becomes 63 against an `inplace_capacity` of 1. That one is quieter and worse: `to_string(x)` returned `0` with no diagnostic, the copy constructor segfaulted, and assigning a fresh value over `x` gave `stack-buffer-overflow` on the write.

Recovery exists only by leaking. `x = big_int{}` and `x = 42` both write into the freed buffer, and retrying the assignment double-frees. `std::move` and `swap` do restore `x`, but they relocate the corrupt control words into the other object, whose destructor then double-frees. Moving into a `big_int` that is never destroyed leaves `x` fully usable at the cost of a leaked 16-byte shell. Every path that eventually runs a destructor over those control words is undefined.

Two notes on scope. The heterogeneous `operator=` at `:429` goes through `grow` and measures strong, so this is specific to the same-type path. And #26 is the direct lineage: its complaint was the unconditional `free_storage()` at the top of the old `operator=`, and the fix added the capacity-reuse fast path but kept the free above the allocation on the slow path.

Possibly relevant to #328, since another throwing path out of assignment would travel this same sequence.

Separately, and much milder: `*=`, `/=` and `%=` are basic-guarantee only at every injection depth I tried, silently leaving the value `0`, sometimes with a large buffer still attached. `<<=`, `+=`, `-=`, the copy constructor, `gcd` and `lcm` were strong at every depth. The paper does not currently state per-operation guarantees, so that may be worth wording rather than code.

xroche:
Investigation and fix were AI-assisted; I ran both debug presets, ASan and UBSan, and `pre-commit` locally.

Closes: #335.

`assign_value`'s slow path freed the destination's buffer and published the source's limb count before allocating, so a throw left the control words describing freed storage. Each branch now writes them only after the storage they describe is in place, and the fresh-allocation branch allocates before it frees, so a throw leaves the destination unchanged.

Under a propagating allocator the new block has to come from the source's allocator while the old one is still released with ours, so it is allocated through a copy of the source's; hence the new static `alloc_limbs_from`. That path had no coverage before, since no allocator in the suite set `propagate_on_container_copy_assignment`, so `allocation.test.cpp` gains a stateful propagating allocator.

### DIFF
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -725,12 +725,10 @@ class BEMAN_BIG_INT_TRIVIAL_ABI basic_big_int {
 
         // Slow path: current buffer is too small. Each branch releases our buffer
         // with our current allocator, before any propagation, and publishes the
-        // control words only once the storage they describe is in place. We may
-        // also be able to adopt src's buffer if the allocators are compatible.
+        // control words only once the storage they describe is in place.
         if (src.is_representation_inplace() && needed <= inplace_capacity) {
-            // Both src and the requested headroom fit inline. Reaching here needs a
-            // heap buffer smaller than the inline one, which nothing produces, so
-            // there is nothing to release; just propagate (if applicable) and copy.
+            // Both src and the requested headroom fit inline. No buffer to
+            // adopt or allocate; just propagate (if applicable) and copy limbs.
             if constexpr (propagate_alloc) {
                 m_alloc = std::forward<Src>(src).m_alloc;
             }
@@ -772,10 +770,9 @@ class BEMAN_BIG_INT_TRIVIAL_ABI basic_big_int {
         }
 
         // Fall back to a fresh allocation of `needed` limbs. Secure it before
-        // releasing our own buffer, so a throwing allocation leaves `*this`
-        // unchanged. When the allocator propagates, the new block has to come
-        // from `src`'s allocator while the old block still has to be released
-        // with ours, so allocate through a copy of `src`'s.
+        // releasing ours, so a throwing allocation leaves `*this` unchanged. A
+        // propagating allocator has to serve the new block while ours still has to
+        // release the old, so allocate through a copy of `src`'s.
         const alloc_result allocation = [&] {
             if constexpr (propagate_alloc) {
                 allocator_type src_alloc(src.m_alloc);
--- a/include/beman/big_int/big_int.hpp
+++ b/include/beman/big_int/big_int.hpp
@@ -576,12 +576,13 @@ class BEMAN_BIG_INT_TRIVIAL_ABI basic_big_int {
     template <detail::cv_unqualified_floating_point F>
     constexpr void assign_from_float(F value) noexcept;
 
-    [[nodiscard]] constexpr alloc_result alloc_limbs(size_type n);
-    constexpr void                       free_limbs(pointer p, size_type n);
-    constexpr void                       free_storage();
-    constexpr void                       grow(size_type limbs_needed);
-    constexpr void                       copy_n_to_allocation(const limb_type* p, size_type n, alloc_result out);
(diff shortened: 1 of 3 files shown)


include/beman/big_int/big_int.hpp
```diff
@@ -722,20 +723,19 @@ class BEMAN_BIG_INT_TRIVIAL_ABI basic_big_int {
...
-        // the allocators are compatible (stateless, propagating, or just equal).
-        free_storage();
-        m_size_and_sign = src.m_size_and_sign;
-
+        // Slow path: current buffer is too small. Each branch releases our buffer
+        // with our current allocator, before any propagation, and publishes the
+        // control words only once the storage they describe is in place. We may
+        // also be able to adopt src's buffer if the allocators are compatible.
         if (src.is_representation_inplace() && needed <= inplace_capacity) {
+            free_storage();
```
eisenwave: This is redundant. There is no storage to be freed if the representation is in-place.

eisenwave (changes requested):
Thanks!

I haven't looked too closely at the tests yet, but the changes to the allocation logic make sense, other than one nitpick.

mborland (commented):
Minus Jan's comment this looks good to me.

include/beman/big_int/big_int.hpp
xroche: Removed, thanks.

The condition tests `src` though, so what makes it safe is that *our* buffer would need to be on the heap with capacity below `inplace_capacity`. I could not find anything that produces one: `grow` doubles from `inplace_capacity`, the fallback allocates above it, `shrink_to_fit` either returns inline or reallocates above it, and the steal branch adopts a capacity already above it. A trap at the top of the branch also left all 969 tests passing. That reads as the branch being dead rather than just not needing the release, but I may be missing a path, so I have left it alone.

mborland (approved):
This looks good to me now.

